PR #13, opened 2017-01-20
Fix #1. Require one variant to be default.

### LINKED ISSUE #1: Require one variant to be the default
#### Goal
Guarantee that a `select-expression` always evaluates to a string.

#### Description
Placeables can contain `select-expression` whose value should always be a string. We can enforce this by requiring one of the variants to be the default using the `*` syntax.

The `member-list` class in the BNF grammar is used both for traits and variants. We will likely want to introduce a new class: `variant-list`.

#### Discussion
https://groups.google.com/forum/#!topic/mozilla.tools.l10n/cpVonPQS0sY

stasm:
Fix #1. This is based on @spagy's pull request in l20n/spec:
https://github.com/l20n/spec/pull/10


stasm:
@zbraniecki Have you had a chance to take a look at this review request?

PR #14, opened 2017-01-20
Fix #2. Disallow multiple expressions in a placeable.

### LINKED ISSUE #2: Disallow multiple expressions in a placeable
#### Goal
Simplify the grammar and the AST.

#### Description
Placeables currently allow more than one expression separated by a comma. The resulting list must be implicitly formatted with a language-specific List formatter before being interpolated into the parent pattern. Separating with a comma conflicts with the planned changes to allowing lists as selectors to `select-expression`.

The implicit list is also a rare construct and at the same time it impacts the shape of the AST for every placeables. I suggest to remove this feature. The `{ LIST(…) }` syntax can be used instead. In the future we can also consider a syntax for list literals but the placeable would still have only one expression.

#### Discussion
https://groups.google.com/forum/#!topic/mozilla.tools.l10n/YBNCSp7J0zU

stasm:
Fixes #2.

### DIFF
--- a/fluent.asdl
+++ b/fluent.asdl
@@ -7,9 +7,7 @@ module Fluent
           | Comment(comment)
           | Junk(string body)
 
-    pat = Pattern(elem* elements, bool quoted)
-    elem = Text(string)
-         | Placeable(expr* expressions)
+    pat = Pattern(expr* elements, bool quoted)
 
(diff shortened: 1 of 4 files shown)


stasm:
@zbraniecki Have you had a chance to take a look at this review request?

PR #19, opened 2017-01-25
Fix #17. Allow newline in arglist

### LINKED ISSUE #17: Allow newline between arguments to a call-expression
After #2 is fixed, we'll run into a similar problem as the one described in https://github.com/l20n/spec/issues/14, but for `arglist`: the current grammar doesn't allow new lines in the list of arguments to a `call-expression`. It makes the complex examples from the syntax guide invalid:

```
liked-photo = { LEN($people) ->
    [1] { $people } likes
    [2] { $people } like
    [3] { LIST(TAKE(2, $people), "one more person") } like

   *[other] { LIST(
        TAKE(2, $people),
        "{ LEN(DROP(2, $people)) ->
            [1]    one more person like
           *[other]  { LEN(DROP(2, $people)) } more people like
        }"
    ) }
} your photo.
```

stasm:


fluent.ebnf
```diff
@@ -43,8 +43,8 @@ expression           ::= quoted-pattern
                       ;
 select-expression    ::= expression __ ' ->' __ variants-list;
 member-expression    ::= identifier '[' keyword ']';
-call-expression      ::= builtin '(' __ arglist? __ ')';
-arglist              ::= argument (__ ',' __ arglist)?;
+call-expression      ::= builtin '(' __ NL? arglist? __ NL? ')';
+arglist              ::= argument (__ NL? ',' __ NL? arglist)?;
```
zbraniecki: Unless I misunderstand ebnf, this will not give you what you want.

In particular, you're saying that call-expression may have any number of white spaces, then optionally one or more NL, then any number of whitespaces, then once more optionally one or more NL and then an argument.

If I'm correct, this means that you can have at maximum two pairs of NL's(one or more) separated by '__' but not three. Seems arbitrary.

Would it make more sense to follow what parser does and do:

```
__ ::= [ \t]*;
___ ::= [ \t\n\r]*;
call-expression ::= '(' ___ arglist? ___ ')';
arglist ::= argument ( ___ arglist)?;
```

stasm: Good catch and I like your suggestion!

stasm: That said, we need a way to _require_ new lines sometimes. I'll work on a new patch later today.

zbraniecki:
```
xx := ___ \n ___;
```

stasm:
This was fixed by #32.

Issue #4, opened 2017-01-04
Allow selectors to be lists


stasm:
#### Goal
Allow for alternative syntax for nested variants.

#### Description
Nested variants are currently hard to write and result in some repetition. ICU's experience with MessageFormat also suggests that it is easier for localizers to work with full sentences as variants.

The proposed solution is to allow lists as selectors in `SelectExpressions` and as variant keys:

```
key = { $items, $buckets ->
  [one, one]     Found 1 item in 1 bucket.
  [one, other]   Found 1 item in { $buckets } buckets.
  [other, one]   Found { $items} items in 1 bucket.
  [other, other] Found { $items} items in { $buckets } buckets.
}
```

#### Dependencies
#2, #3.

#### Discussion
https://groups.google.com/forum/#!topic/mozilla.tools.l10n/YBNCSp7J0zU

stasm:
I've removed this issue from the 0.3 milestone. I'd like to release 0.3 shortly after 0.2 and I don't want to block it with this issue. It's something that we all agree is desirable but it's also pretty big in scope.  As the proposal is backwards-compatible, we can add it in a future milestone—when we're ready.

gunchleoc:
I agree that having full sentences is better, but that will only be workable if you need only a few of the rules - it's a lot of typing otherwise. So, as an option, great, but it should not be made mandatory. e.g. for gd we would get 4x4 = 16 entries for such a string.

stasm:
The last time I talked to @zbraniecki about this, we considered the following syntax for list-typed variant keys:

```properties
key = { $items, $buckets ->
    [one][one] Found 1 item in 1 bucket.
    [one][other] Found 1 item in { $buckets } buckets.
    [other][one] Found { $items} items in 1 bucket.
   *[other][other] Found { $items} items in { $buckets } buckets.
}
```

It would be similar, but not equivalent, to the exploded version:

```properties
key = { $items ->
    [one] { $buckets ->
        [one] Found 1 item in 1 bucket.
       *[other] Found 1 item in { $buckets } buckets.
    }
   *[other] { $buckets ->
        [one] Found { $items} items in 1 bucket.
       *[other] Found { $items} items in { $buckets } buckets.
    }
}
```

The key difference between the two is that in the first one, `[other][other]` is the default value for any errors in either `$items` or `$buckets` while in the second one, `$items` can properly resolve to `[one]` if the only error is in `$buckets`.

[...]

PR #21, opened 2017-01-26
Fix #3. Use { and } for explicit grouping

### LINKED ISSUE #3: Explicit precedence of expressions
#### Goal
Give users more control when working with lists.

#### Description
We want to allow list-typed selectors in `select-expression` which would lead to ambiguous syntax in constructs like `{ LIST(foo, bar, baz -> … )}`.  A work-around currently exists:

`{ LIST(foo, "{bar, baz -> … }")}`

I suggest we introduce explicit syntax for grouping which will instruct the parser unambiguously about the order of expressions.  The proposed syntax is `{ … }` for grouping:

`{ LIST(foo, {bar, baz -> … })}`

#### Discussion
https://groups.google.com/forum/#!topic/mozilla.tools.l10n/YBNCSp7J0zU

stasm:
This only changes the EBNF grammar and doesn't change the ASDL. The grouping is supposed to help the parser in potentially ambiguous situations. Once the parser figures out the parse tree, the final AST will be unambiguous.

### DIFF
--- a/fluent.ebnf
+++ b/fluent.ebnf
@@ -32,15 +32,17 @@ quoted-text          ::= ([^{"] | '\{' | '\"')+
 block-text           ::= NL __ '|' unquoted-pattern
 
 placeable            ::= '{' __ expression __ '}'
-expression           ::= quoted-pattern
+expression           ::= selector-expression | select-expression
+
+selector-expression  ::= quoted-pattern
                        | number
                        | identifier
                        | variable
-                       | select-expression
                        | member-expression
                        | call-expression
+                       | placeable
 
-select-expression    ::= expression __ ' ->' __ variants-list
+select-expression    ::= selector-expression __ ' ->' __ variants-list
 member-expression    ::= identifier '[' keyword ']'
 call-expression      ::= builtin '(' __ arglist? __ ')'
 arglist              ::= argument (__ ',' __ arglist)?
--- a/fluent.asdl
+++ b/fluent.asdl
@@ -9,14 +9,16 @@ module Fluent
 
     pat = Pattern(expr* elements, bool quoted)
 
-    expr = MessageReference(iden id)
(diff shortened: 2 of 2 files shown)


fluent.ebnf
```diff
@@ -32,15 +32,17 @@ quoted-text          ::= ([^{"] | '\{' | '\"')+
...
+expression           ::= selector-expression | select-expression
+
+selector-expression  ::= quoted-pattern
                        | number
                        | identifier
                        | variable
-                       | select-expression
                        | member-expression
                        | call-expression
+                       | placeable
```
zbraniecki: why placeable?

```
key = { { $foo } ->
  [one] Value
}
```

is that intended to work?

stasm: Yes. It's not what the serialize would normally serialize to, but I don't think we need to artificially limit the syntax here. It doesn't change the meaning compared to just using `$foo`, just like `(1) === 1` in JavaScript.

zbraniecki: So, what's the use case?

stasm: The use case is tightly related to #4. In particular, we need a way to remove the ambiguity from `call-expressions` with multiple arguments, some of them being `select-expressions`:

```
key = { LIST(foo, { bar, baz -> 
        [bar, baz] Bar and Baz
    }) }
```

stasm:
Thanks for the review. After talking to @Pike last Friday I partly changed my mind about ASDL. The `placeable` type shouldn't be part of the ASDL, still, but I think it makes sense to separate the following two concepts:

- expressions which can be selectors in a `select-expression`,
- expressions which can be inside of placeables.

I updated the PR to reflect that. It should map closely to the data structures we'll have in the Rust implementation. @zbraniecki, this probably doesn't need another review from you, but I wanted to let you know before I merge.

PR #23, opened 2017-01-30
Fix #5. Allow select-expression with an empty selector

### LINKED ISSUE #5: Allow select-expression with an empty selector
#### Goal
Define variants of a message which can be accessed from other strings.

#### Description
We currently define variants of the message which are local and specific to the natural language using traits. For instance, we'd write:

```
brand-name =
   *[nominative] Firefox
    [genitive]    Firefox's
```

This is a bit weird because in the above example, "Firefox" is the real value of the message and nominative and genitive are the value's facets.  However the `value` field of the message remains empty due to the fact that the facets are defined as traits.  These traits also need to remain private which in turn necessitates a way to mark public traits as public; we currently use namespaces for this purpose.

The proposed change is to allow defining facets of a value as variants of a `select-expression` which doesn't have a selector:

```
brand-name = {
   *[nominative] Firefox
    [genitive]    Firefox's
}
```

This would also allow to remove namespaces from member keys (keywords).  The `member-expression` would change the meaning: it would now be used to access a named variant of the first placeable in the value of the message. If that placeable isn't a `select-expression` or already has a selector, normal evaluation would follow.

```
about1 = About { brand-name }
about2 = About { brand-name[nominative] }
```

#### Dependencies
#1.

#### Discussion

stasm:


fluent.ebnf
```diff
@@ -32,7 +32,7 @@ quoted-text          ::= ([^{"] | '\{' | '\"')+
 block-text           ::= NL __ '|' unquoted-pattern
 
 placeable            ::= '{' __ expression __ '}'
-expression           ::= selector-expression | select-expression
+expression           ::= selector-expression | select-expression | variants-list
```
zbraniecki: you need to add `variant-list`

stasm: it's... right there. in the line you commented on.

zbraniecki: the reference to it is, but the definition is nowhere to be found. Did I miss something?

stasm: ah :) it's defined in line 23; not part of this patch.

zbraniecki: somehow github hid it from me

Issue #7, opened 2017-01-04
New syntax for meta-data


stasm:
#### Goal
Provide a simple means for defining private meta-data for messages.

#### Description
Currently, meta-data can be added to messages by using traits. Traits without namespaces are considered private.

```
brand-name =Firefox
  [gender] masculine
```

#5 and #6 will simplify traits and we'll need a new way to encode meta-data.

The proposal is to use binary tags attached to the value:

```
#masculine
brand-name = Firefox
```

The benefit of the binary approach is that there's usually no need to name the property in question (_gender_).

#### Discussion
https://groups.google.com/forum/#!topic/mozilla.tools.l10n/dhWfBXHzuZI

Pike:
One of the big wins of FTL was that everything about the message was in the value part of the syntax.

There is some level of beauty that you start the message with the ID. Right now, only comments break that. For tooling that is a boost.

Also, does the 

    #masculine vs
    # masculine

create challenges in error recovery? The ' ' would also be a typo that would be really hard to debug, as

    # masculine
    #masculine
    bob = Bob

would be totally legal, right?

zbraniecki:
I think my vote would go for semantic comments for meta-data.

stasm:
I think we used the `#` sigil as an example and then I copied it here without realizing we already use `#` for comments!

I'd love to discuss about semantic comments more. JSDoc-style `@param` clauses would certainly help tooling. And it might be possible to encode language-specific meta-information in the comments as well (`@meta foo`?).

Pike:
@phlax, @ta2-1, does prefixing messages with metadata impact how pootle implements l20n support?

ta2-1:
Hi @Pike, thanks for heads up. I'll check.

ta2-1:
@Pike, I’d say that we use l20n libraries so as long as that is consistent we’re pretty much unimpacted by syntax changes. I believe that @mathjazz is in the same situation.

stasm:
@Pike suggested that we separate the meta-information from semantic comments. The reason for this is that he sees semantic comments as relating to the toolchain and the process (`@param`, `@rev`), while the meta-information is strictly language-specific and private.

He suggested the following syntax:

```
# The short name of the app. 
brand-name = {
       *[nominative] Firefox
        [genitive] Firefox's
    }

    [masculine, inanimate]
```

The reason to use the brackets is that it closely resembles the way this information will be used. This in turn improves the copy&paste-ability of the syntax:

```
[...]

PR #24, opened 2017-01-30
Fix #6. New syntax for attributes

### LINKED ISSUE #6: Rename traits to attributes and change syntax
#### Goal
Provide easier syntax for multi-valued messages.

#### Description
While #5 is about multiple _facets_ of the same value, this is about multiple values. This is useful in case of web components or other UI widgets.

Currently we'd write:

```
key = Value
  [label] Label
```

This looks similar to a variant of a `select-expression` but actually defines a trait and "Label" is a separate value, one of two values of the `key` message.  When #1 is fixed this will become even more confusing, because lists of traits will not require defaults, but will use the same syntax as lists of variants.

The proposed solution is to use `.` (the dot) as a new syntax for defining traits. Also, to better reflect their purpose, we want to rename traits to _attributes_.

```
key = Value
key.label = Label
```

If there is no value it's definition can be omitted:

```
file-open.label = Open File
file-open.accesskey = O 
```

This will also require an addition of `attribute-expression`:

```
attribute-expression ::= identifier '.' keyword;
```

```
file-open.label = Open File
file-open.accesskey = O 
file-open.title = To open an existing file press Ctrl + { file-open.accesskey }.
```

#### Discussion

stasm:
This implements the following proposal from #6 for the syntax for public attributes:

```
file-open
  .label     = File Open
  .accesskey = O 
```

I also removed namespaces from variant keys. Attributes are public. The API returns the value and all attributes for a given identifier.

I removed the section on storing private grammatical information from the guide. I'll bring it back in #7.

### DIFF
--- a/fluent.ebnf
+++ b/fluent.ebnf
@@ -10,19 +10,21 @@ char                 ::= [https://www.w3.org/TR/REC-xml/#NT-Char]
 __                   ::= [#x20#x9]*   /* space, tab */
 NL                   ::= [#xA#xD]+    /* line feed, carriage return */
 
-identifier           ::= [a-zA-Z_.?-] ([a-zA-Z0-9_.?-])*
+identifier           ::= [a-zA-Z_?-] ([a-zA-Z0-9_?-])*
 variable             ::= '$' identifier
 keyword              ::= [a-zA-Z_.?-] ([a-zA-Z0-9_.?- ]* [a-zA-Z0-9_.?-])?
 builtin              ::= [A-Z_.?-]+
 number               ::= [0-9]+ ('.' [0-9]+)?
 
-member-key           ::= number | (identifier '/')? keyword
-member               ::= '[' member-key ']' __ pattern NL
-default-member       ::= '*' member
-variants-list        ::= NL (__ member)* __ default-member (__ member)*
-traits-list          ::= variants-list | NL (__ member)+
+variant-key          ::= number | keyword
+variant              ::= '[' variant-key ']' __ pattern NL
+default-variant      ::= '*' variant
+variant-list         ::= NL (__ variant)* __ default-variant (__ variant)*
 
-message              ::= identifier __ '=' __ (pattern | pattern traits-list | traits-list)
+attribute            ::= '.' identifier __ '=' __ pattern NL
+attribute-list       ::= NL (__ attribute)+
+
+message              ::= identifier __ '=' __ (pattern attribute-list? | attribute-list)
 pattern              ::= unquoted-pattern
                        | quoted-pattern
 unquoted-pattern     ::= (unquoted-text | placeable | block-text)+
@@ -38,11 +40,13 @@ selector-expression  ::= quoted-pattern
                        | number
                        | identifier
                        | variable
-                       | member-expression
+                       | attribute-expression
+                       | variant-expression
                        | call-expression
                        | placeable
 
-select-expression    ::= selector-expression __ ' ->' __ variants-list
+select-expression    ::= selector-expression __ ' ->' __ variant-list
+attribute-expression ::= identifier '.' identifier
 variant-expression   ::= identifier '[' keyword ']'
 call-expression      ::= builtin '(' __ arglist? __ ')'
 arglist              ::= argument (__ ',' __ arglist)?
--- a/guide.rst
+++ b/guide.rst
@@ -259,41 +259,19 @@ without a selector and referred to from other messages, like the ``about``
 message above.
 
 
-Storing Additional Information
-==============================
-
-::
-
-    brand-name = Firefox
-        [gender] masculine
-
-    opened-new-window = { brand-name[gender] ->
-       *[masculine] { brand-name } otworzyl nowe okno.
-        [feminine] { brand-name } otworzyla nowe okno.
-    }
-
-Traits can be used to describe parameters of the entity that can be then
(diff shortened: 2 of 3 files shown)


fluent.ebnf
```diff
@@ -16,13 +16,15 @@ keyword              ::= [a-zA-Z_.?-] ([a-zA-Z0-9_.?- ]* [a-zA-Z0-9_.?-])?
...
+variant-key          ::= number | keyword
+variant              ::= '[' variant-key ']' __ pattern NL
+default-variant      ::= '*' variant
+variant-list         ::= NL (__ variant)* __ default-variant (__ variant)*
 
-message              ::= identifier __ '=' __ (pattern | pattern traits-list | traits-list)
+attribute            ::= '.' identifier __ '=' __ pattern NL
+attribute-list       ::= NL (__ attribute)+
+
+message              ::= identifier __ '=' __ (pattern | pattern attribute-list | attribute-list)
```
zbraniecki: can you write it as:
```
message              ::= identifier __ '=' __ (pattern attribute-list? | attribute-list)
```

fluent.ebnf
```diff
@@ -16,13 +16,15 @@ keyword              ::= [a-zA-Z_.?-] ([a-zA-Z0-9_.?- ]* [a-zA-Z0-9_.?-])?
...
-default-member       ::= '*' member
-variants-list        ::= NL (__ member)* __ default-member (__ member)*
-traits-list          ::= variants-list | NL (__ member)+
+variant-key          ::= number | keyword
+variant              ::= '[' variant-key ']' __ pattern NL
+default-variant      ::= '*' variant
+variant-list         ::= NL (__ variant)* __ default-variant (__ variant)*
 
-message              ::= identifier __ '=' __ (pattern | pattern traits-list | traits-list)
+attribute            ::= '.' identifier __ '=' __ pattern NL
```
zbraniecki: the way you use NL indicates that EOF should be part of NL. Like here we will recognize the end of the attribute by either EOF or NL.

guide.rst
```diff
@@ -259,41 +259,20 @@ without a selector and referred to from other messages, like the ``about``
...
 
 ::
 
     login-input = Predefined value
-        [html/placeholder] example@email.com
-        [html/aria-label]  Login input value
-        [html/title]       Type your login email
+        .placeholder = example@email.com
+        .aria-label = Login input value
+        .title = Type your login email
```
zbraniecki: would it make sense to keep the values it aligned?

fluent.ebnf
zbraniecki: should message's pattern also end with NL? (like variant's?)

fluent.ebnf
stasm: Let's fix this in 0.3 together with other white-space related bugs. For now I wrote it similar to how `variant` is defined.

fluent.ebnf
stasm: `variant-list` starts with `NL`, so that's covered.

guide.rst
stasm: I remember talking to @flodolo about this: aligning values often times ends up creating diff noise and his recommendation was not to do it. And it will also make it easier to write a syntax-formatting tool for FTL that way :)

guide.rst
```diff
@@ -259,41 +259,20 @@ without a selector and referred to from other messages, like the ``about``
...
-used in other selectors.
-
-Imagine an entity ``brand-name`` that can be either *Firefox* or *Aurora*.  The 
-former is *masculine*, while the latter is *feminine*, so sentences that refer 
-to this entity may want to branch depending on the gender of it.
-
-
-HTML/XUL Attributes
-===================
+HTML Attributes
```
Pike: Can we call this DOM?

guide.rst
Pike: flod is generally annoyed by aligning to keys as you completely break blame each time you modify one of the entries.

I think we should recommend the practice to not align one attribute to others.

fluent.ebnf
zbraniecki: is it? If the message has only key + pattern, then is NL at the end of line part of pattern or not? Right now it's not if the pattern is on message, but it is if it's on an attribute.

stasm: Each `entry` ends with `NL` too, so I think we end up with an new line at the end in all scenarios.

guide.rst
flodolo: Scenario is: new key, longer than existing one, you end up with a huge diff just to align separators. Add to that that most diffs are not as smart as GitHub's diff logic.

zbraniecki: so, I just produced a new test FTL file using the new syntax. It's the closest thing we have from Firefox that resembles "big" FTL file and as such I believe it's valuable for us to test our syntax on it against our principles.

To remind principles:
 - the file should be readable without prior knowledge of FTL
 - the file should be easy to scan through
 - the file should be editable with minimal FTL knowledge

0.1 syntax: https://github.com/projectfluent/fluent-rs/blob/0.1/tests/workload-low.ftl
0.2 syntax: https://github.com/projectfluent/fluent-rs/blob/master/tests/workload-low.ftl

For one, I'd say we did manage to make it look cleaner (less sigils!)
But also, I believe the attribute alignment significantly helps with readability, so I'm not sure if I agree that preserving blame is worth it.

PR #26, opened 2017-02-01
Simplify Expressions

stasm:
This simplifies the ASDL by merging `elem`, the explicit `plcbl` type and
`expr` into a single `expr` type.

The previous version was inspired directly by the EBNF grammar and reflected
the grouping of expressions achieved by using additional braces pairs right in
the AST.  The purpose of the grouping is to help the parser disamiguate the
syntax.  The produced AST however, by its tree-like structure, is never
ambiguous with regard to operator precedence.

A more explicit AST can help serialization: upon seeing a Placeable data type
the serializer knows to serialize it wrapped in braces.  However, the
serializer will still need to check the parents of Patterns to decide if they
should be wrapped in quotes or not.  The same technique can be used when
serializing Expressions without sacrifices to the type-strictness of the AST.

The resulting AST is simpler than the previous one.

### DIFF
--- a/fluent.asdl
+++ b/fluent.asdl
@@ -8,24 +8,19 @@ module Fluent
           | Junk(string body)
 
     -- Pattern values
-    pat = Pattern(elem* elements, bool quoted)
-    elem = String(string)
-         | Placeable(plcbl)
+    pat = Pattern(expr* elements, bool quoted)
 
-    -- Expressions allowed inside of braced Placeables
-    plcbl = Expression(expr)
-          | SelectExpression(expr? expr, var* vars)
-
-    -- Expressions allowed as selectors of SelectExpression and
-    -- arguments to CallExpression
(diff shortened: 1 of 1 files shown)


stasm:
@zbraniecki, can you take a look? The ASDL described the data structure that a parser should produce for consumption by the resolver. In this PR I diverged from a very close mapping to EBNF on purpose: while the EBNF can be more specific in certain cases to help the parser disambiguate or forbid code which is too complex, the AST can be a bit more relaxed. This should make it easier to implement the resolver.

The type system in languages like Rust will still enforce correctness of the syntax after these changes. A hand-built AST which meets the type checking criteria can then be serialized to a valid FTL file. When serializing `Expression` and `Patern` nodes we will need to check their parents to decide if they should be wrapped in braces and quotes respectively.

PR #31, opened 2017-02-13
Fix #30. Allow more characters in keys, and trim whitespace around them

### LINKED ISSUE #30: Allow more characters in keys, and trim whitespace around them
Variant keys can be language-specific and as such should accept a wide range of Unicode characters. For example, this will allow localizers to define grammatical cases in their native language.

Let's also make them more liberal on the white-space around them and always trim on both ends.

stasm:


stasm:
Thanks for the review, @Pike. I introduced `keyname` separate from `keyword` specifically in preparation for tags. Currently, in a variant `[foo bar]`, the entire `foo bar` part is called a key _word_ which I find confusing because it's two words, obviously. So in the PR I'd like to call `foo bar` a key _name_ consisting of two key _words_.  Tags will consist of a single `keyword`.

Let me know if you'd prefer to see different naming here; I picked `keyword` and `keyname` mostly because I didn't have better ideas.  They are less strict than `identifiers` but stricter than just `text`.

stasm:
@Pike does my comment https://github.com/projectfluent/syntax/pull/31#issuecomment-279528392 address your concerns?

Pike:
I'm still a bit clueless as to the use case of a whitespace within variant-keys, aka [ foo bar ]. Maybe for literal value matches? I know, not a regression, but this makes me look.

I wonder if just "word" instead of "keyword" would be better? At least I have less assumptions on it.

I think we should make section labels refer to something else than variant-keys. Maybe just word (__ word)* directly? I find the shared grammar production between the two rather confusing.

stasm:
@zbraniecki any thoughts on limiting `variant-key` to single `words` for now?

stasm:
In the latest iteration I went for `word` and `symbol` as a collection of words.

zbraniecki:
No strong opinion.

My mental use case for whitespace in variant was sth like:
```
city = { $city ->
  *[default] { $city}
   [Warsaw] Warszawa
   [New York] Nowy Jork
}
```

but I think we can add it later if needed.

stasm:
I think I'd prefer no to limit this artificially. Since the brackets provide a good hint about where the key name ends, I think we can allow spaces inside.

Issue #22, opened 2017-01-27
Define the behavior of backslash


stasm:
In https://github.com/projectfluent/syntax/issues/12#issuecomment-275678121 I said we'd need to define the exact behavior of the backslash character `\` for the purposes of escaping. This includes defining:

- the list of known escape sequences (`\ ` ( a space), `\t`, `\n`, `\*`, `\[`, `\{`, `\u`, `\\`, others?),
- how the Unicode escapes work: is `\u20` valid and the same as `\u0020`?
- the behavior of unknown sequences, like `\a` (does the backslash take the following character out of the syntax parsing?),
- the behavior for edge-cases, like:

        foo\bar = Foobar

  Is that a syntax error? If not, what is the name of the identifier?

        foo = Foo\
        bar = Bar

  Is that an escaped new-line?

stasm:
A draft proposal:

  - Escape sequences are only allowed in the `text` and `quoted-text` productions.

  - Known escapes are: `\\`, `\*`, `\[`, `\{`, `\uXXXX`, `\t`, `\n`, `\"`.

      - Do we actually need `\t` and `\n`? The whole syntax was designed to make it easy to use white-space and those characters could be written literally, if needed.

  -  Unicode sequences are only valid with four characters: `\u0020` is valid, `\u20` is not.  This is the same as in JavaScript.

      - ES2015 introduced [Unicode code point escapes](http://ecma-international.org/ecma-262/6.0/#sec-literals-string-literals) which are written as `\u{XXXXXXX}` with any number of `X`, thus allowing representing code points from outside of the Basic Multilingual Plane without resorting to using surrogate pairs. Non-BMP code points are very rare but the questions still remains: are we okay with using surrogate pairs for them if we go with the `\uXXXX` proposal?

  - Escaping any other character returns the character itself and the character is parsed as normal; `\a` results in `a` and `\` at the end of line results in the EOL character which is parsed as normal. This is different from JavaScript which has a special case for `\EOL` which is called `LineContinuation` and technically is not an escape sequence.

    For instance, in the following example, the escaped EOL results in a real EOL and it ends the value part of the `a` variant:

    ```
    foo = {
           *[a] AAA\
            [b] BBB
        }
    ```

@Pike, @zbraniecki — I'd love to hear your thoughts on this. Thanks!

zbraniecki:
sgtm! I'd not do `\n`, `\t` until we have a use case.

stasm:
[...]

Issue #18, opened 2017-01-25
incomplete unquoted-text definition


GlenDC:
Currently the definition of an unquoted-text is as follows:

```EBNF
unquoted-text        ::= ([^{] | '\{')+;
```

source: https://github.com/projectfluent/syntax/blob/master/fluent.ebnf#L28

For most cases this is fine, however when just relying on this definition it breaks in following example:

```FTL
opened-new-window = { brandName[gender] ->
 *[masculine] { brandName } otworzyl nowe okno.
  [feminine] { brandName } otworzyla nowe okno. }
```

In the example given above, it will read the entire line, including the '}' character, which is actually the closing bracket of the root `placeable`. This problem is fixed if you define an `unquoted-text` as follows:

```EBNF
unquoted-text        ::= ([^{}] | '\{' | '\}')+;
```

stasm:
Thanks for the report! `member` is currently defined with the final new line on its end. I've created the 0.3 milestone to focus on the white spacing rules in the syntax. I'll come back to this issue then.

stasm:
This was fixed by #32. We changed the syntax of multiline strings by removing the leading `|` and requiring indentation.

PR #32, opened 2017-02-15
Require message body to be indented

### LINKED ISSUE #12: Enforce indentation for multiline string and variants
Currently, this form is allowed:

```rust
key =
|Value
```

I think it's likely to be confusing and in the spirit of "it's easier to relax than tighten", I'd like to suggest enforcing a requirement for at least a single whitespace before `|` character.

If we also do the same for selector variants to prevent:

```rust
key = { $OS ->
[one] Foo { $OS2 ->
[two Faa
}
}
```

then we'll make it easier to visually spot new entries because the only characters at the beginning of the new line will be `#`, `[[`, `ident_start` and `}` which is easy to scan through in a big file.

stasm:
Fix #12, #17, #18.

With this change, the entire body of a message needs to indented. This makes
error recovery very easy: finding the next message definition is as simple as
finding the next identifier with no indentation.

It also opens up a number of opportunities: we can remove the `|` syntax for
multiline blocks of text and allow line breaks inside of placeables safely.

The PR also allows the value to be defined on a new line, making the
following examples equivalent:

    lipsum = Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi
        pellentesque congue metus, non mattis sem faucibus sit amet.

    lipsum
        = Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi
        pellentesque congue metus, non mattis sem faucibus sit amet.

I hope this will help when attributes are present:

    lipsum
        = Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi
        pellentesque congue metus, non mattis sem faucibus sit amet.

        .attr = Attribute

Lastly, quoted patterns are only available inside of placeables and cannot be
used directly as values.

The exact semantics of `\` escapes will be defined in #22.

### DIFF
--- a/guide/builtins.md
+++ b/guide/builtins.md
@@ -5,7 +5,7 @@ emails = You have { $unreadEmails } unread emails.
 emails2 = You have { NUMBER($unreadEmails) } unread emails.
 
 last-notice =
-    | Last checked: { DATETIME($lastChecked, day: "numeric", month: "long") }.
+    Last checked: { DATETIME($lastChecked, day: "numeric", month: "long") }.
 ```
 
 ```json
--- a/guide/functions.md
+++ b/guide/functions.md
@@ -126,16 +126,16 @@ the purpose of choosing the correct plural category:
 
 ```
 liked-count = { $num ->
-  [0]     No likes yet.
-  [one]   One person liked your message
- *[other] { $num } people liked your message
-}
+        [0]     No likes yet.
+        [one]   One person liked your message
+       *[other] { $num } people liked your message
+    }
 
 liked-count2 = { NUMBER($num) ->
-  [0]     No likes yet.
-  [one]   One person liked your message
- *[other] { $num } people liked your message
-}
+        [0]     No likes yet.
+        [one]   One person liked your message
+       *[other] { $num } people liked your message
+    }
 ```
 
 ### `DATETIME`  
--- a/guide/selectors.md
+++ b/guide/selectors.md
@@ -2,9 +2,9 @@
 
 ```
 emails = { $unreadEmails ->
-    [one] You have one unread email.
-   *[other] You have { $unreadEmails } unread emails.
-}
+        [one] You have one unread email.
+       *[other] You have { $unreadEmails } unread emails.
+    }
 ```
 
 ```json
@@ -34,7 +34,7 @@ unformatted number:
 
 ```
 your-score = { NUMBER($score, minimumFractionDigits: 1) ->
-    [0.0]   You scored zero points. What happened?
-   *[other] You scored { NUMBER($score, minimumFractionDigits: 1) } points.
-}
+        [0.0]   You scored zero points. What happened?
+       *[other] You scored { NUMBER($score, minimumFractionDigits: 1) } points.
+    }
 ```
--- a/guide/text.md
+++ b/guide/text.md
@@ -1,23 +1,16 @@
-# Working With Text: Multiline and Quotes
+# Working With Text
 
 ```
 about = About Our Software
 description =
-    | Loki is a simple micro-blogging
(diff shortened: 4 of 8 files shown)


stasm:
@Pike, @zbraniecki -- this is an early version of the change I'd like to propose. @Pike will recognize ideas from our brainstorming session a few week ago.

I think there are still some things missing here. One is handling completely blank lines. I'd like to get your feedback at this stage. Let me know what you think, thanks!

zbraniecki:
> Lastly, quoted patterns are only available inside of placeables and cannot be
used directly as values.

Does it remove the ability to use quoted patterns in order to get significant trailing/leading white spaces?

stasm:
> Does it remove the ability to use quoted patterns in order to get significant trailing/leading white spaces?

Just a little bit. In #22 I'd like to define all backslash-escapes and allow escaping leading and trailing whitespace either explicitly or through Unicode escapes.

Do you think it's a common enough use-case that we should give it better support?

zbraniecki:
Idk, we can ask @flod. But I know that from the readability principle perspective:

```
key = "  Value"
```

is a very clear way to say - I want `key` to have this value and those two whitespaces in front are meaningful.
Using unicode escapes or backslashes to annotate the same is a departure from that principle.

Pike:
I'll look into the actual diff here in more detail, but regarding the leading/trailing whitespace, just one \ should suffice?

    key = \     Value

would easily denote "     Value".

Trailing whitespace is more tricky, as it's hard to distinguish "\" from "\ ".

That said, on this particular aspect, my experience is that we rarely see leading or trailing whitespace in strings. In particular, if we would, we might want to make whitespace around `=` significant, wouldn't we?

zbraniecki:
is there any particular reason to remove the quote-delimited strings?

If it's just for simplicity, we could ensure we don't lock ourselves out of this solution and keep it in a sleeve in case a use case pops up later.

stasm:
I would be OK forbidding quoted strings in placeables: `{ "foo" }` as well as changing the syntax of `name-argument` to only allow a single unquoted `word` as value.

flodolo:
While I agree that `key = "  Value"` is really clear, it's also a bit confusing as a localizer looking directly at the FTL file. 

First thought when put in front of this: are these quotes going to show up in the product? Should I replace these with prettier/proper double quotes?

If the expectation is for localizers to work in tools, and for those who don't to be smart enough to be familiar with the syntax, I think that's a good choice.

Escaping the leading whitespace seems more confusing to me. Also reminds me of multiline properties, even if it's leading and not trailing.

P.S. sadly flod was taken when I signed up to GH, so you pinged a random inactive user, but I'm watching this project these days

Pike:
On the actual diff, there are a few edgecases still.

AFAICT, there can be one empty line, but only one at a time in a message body?

```
key = value

  .attr = attribute
```
is ok, but
```
key  = value


 .attr = attribute
```
isn't?

Also, there are a few cases where EOF is probably OK instead of NL?

I personally would also like to see
```
key = 
    Some lengthy content here that doesn't have a = at the
    beginning.
```

While thinking of EOF, I also noticed that the grammar doesn't allow for leading whitespace-only lines.

stasm:
I updated the PR to address @Pike's feedback. (Thanks for the help!)

I kept the quoted strings, but with a change to their semantics: they can only be _simple_ strings right now which means they must not contain any placeables.  They can be used as values to `named-arguments` and as literals in placeables.

The latter is useful for providing a way to unambiguously encode leading and trailing white-space into a value:

    foo = { "" }       Foo       { "" }

Leading and trailing white-space will be vary rare—in fact we should actively discourage it. This approach makes it _possible_ but doesn't optimize the syntax for it in any way.

PR #34, opened 2017-02-16
Fix styling nits and some typos across the project

flodolo:
Since you’re using third level titles in some docs, I think it makes sense to stick to the hash syntax.

Also, use code blocks and not blockquotes for code, h1 for the main title of the doc instead of h2


stasm:
Thanks, @flodolo, it reads much better now!

PR #35, opened 2017-02-23
Fix #7. Introduce tags

### LINKED ISSUE #7: New syntax for meta-data

stasm:
Tags are binary values attached to messages.  They are language-specific and
can be used to describe grammatical characteristics of the message.

    brand-name = Firefox
        #masculine

    brand-name = Aurora
        #feminine
        #vowel

Tags can be used in select expressions by matching a hashtag name to the
message:

    has-updated = { brand-name ->
        [masculine] …
        [feminine] …
       *[other] …
    }

Tags can only be defined on messages which have a value and don't have any
attributes.

This PR depends on #32.

### DIFF
--- a/guide/SUMMARY.md
+++ b/guide/SUMMARY.md
@@ -6,6 +6,7 @@
 * [External Data](external.md)
 * [Selectors](selectors.md)
 * [Variants](variants.md)
+* [Tags](tags.md)
 * [Attributes](attributes.md)
 * [Sections](sections.md)
 * [Comments](comments.md)
--- a/guide/tags.md
+++ b/guide/tags.md
@@ -1,0 +1,40 @@
+# Tags
+
+```
+brand-name = Aurora
+    #żeński
+
+has-updated = { brand-name ->
+        [męski] { brand-name} został zaktualizowany.
+        [żeński] { brand-name } została zaktualizowana.
+       *[inny] Program { brand-name } został zaktualizowany.
+    }
+```
+
+Sometimes translations might vary depending on some grammatical trait of
+another message.  In the example above the form of the past tense of _has been
+updated_ depends on the grammatical gender of `brand-name`.
+
+In such cases you can _tag_ messages with simple one-word _hashtags_ and then
+define different translations based on these tags.  You define them with `#`
+which must start on a new line under the message, indented.
+
+Hashtags are specific to your language's grammar and don't have to be in
+English. In the example above, _żeński_ means _feminine_, _męski_ is
+_masculine_ and _inny_ is _other_.
+
+Both tags and variants discussed in the previous chapter can be used to provide
+more information about a message that is specific to your language.  There
+a number of differences between them, however.
+
+  - Tags don't have a value; they _are_ the value.
+
+  - Tags can only be used for matching.  They cannot be inserted into another
+    translation.
+
+  - Tags describe grammatical traits; they should answer the question _Is this
+    message &lt;tagname&gt;_?  For instance, _Is this message feminine?_
+
+  - On the other hand, variants define different _facets_ of the message.  It's
+    the same value, just in slightly different forms to make it grammatically
+    correct when used inside of other messages.
--- a/spec/CHANGELOG.md
+++ b/spec/CHANGELOG.md
@@ -2,6 +2,30 @@
 
 ## Unreleased
 
+  - Introduce tags for language-specific grammatical information.
+
+    Tags are binary values attached to messages.  They are language-specific and
+    can be used to describe grammatical characteristics of the message.
+
+        brand-name = Firefox
+            #masculine
+
+        brand-name = Aurora
+            #feminine
+            #vowel
+
+    Tags can be used in select expressions by matching a hashtag name to the
+    message:
+
+        has-updated = { brand-name ->
+                [masculine] …
+                [feminine] …
+               *[other] …
+            }
+
+    Tags can only be defined on messages which have a value and don't have any
+    attributes.
+
   - Require the message body to be indented.
 
     Quoted strings are now only valid in placeables and cannot contain other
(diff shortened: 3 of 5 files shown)


syntax/fluent.ebnf
```diff
@@ -1,41 +1,53 @@
...
+
+/* special chars: { } [ \ ] */
+special              ::= [#x5b#x5c#x5d#x7b#x7d]
+
+/* line feed, carriage return; space, tab */
+line-break           ::= [#xA#xD]+
+inline-space         ::= [#x20#x9]+
+
+/* if you break lines you need to indent afterwards */
+break-indent         ::= (line-break* inline-space)+
```
zbraniecki: This seems to forbid the edge case where you do "\n\n  Indented test" right?

syntax/fluent.ebnf
```diff
@@ -1,41 +1,53 @@
...
+
+_                    ::= inline-space
+NL                   ::= line-break
+__                   ::= break-indent
```
zbraniecki: what's the reason behind defining two keys for `inline-space` and two for `break-indent`?

syntax/fluent.ebnf
```diff
@@ -1,41 +1,53 @@
...
-unquoted-pattern     ::= (unquoted-text | placeable | block-text)+
-quoted-pattern       ::= '"' (quoted-text | placeable)* '"'
-unquoted-text        ::= ([^{] | '\{')+
-quoted-text          ::= ([^{"] | '\{' | '\"')+
-block-text           ::= NL __ '|' unquoted-pattern
-
-placeable            ::= '{' __ (expression | select-expression | variant-list) __ '}'
-expression           ::= quoted-pattern
+variant-symbol       ::= word (_ word)*
+variant              ::= NL __ '[' _ variant-key _ ']' __ pattern
```
zbraniecki: so, variants are required not to have white spaces?
This will make 
```
key = { $foo ->
  *[one] One
}
```

not work

syntax/fluent.ebnf
```diff
@@ -1,41 +1,53 @@
...
+default-variant      ::= NL __ '*[' _ variant-key _ ']' __ pattern
+variant-list         ::= variant* default-variant variant*
+
+tag                  ::= '#' word
```
zbraniecki: I'm a bit concerned about the ability to put any character here. The idea that:
```
key = Value
  #####
  #!
  #@
  #⁂
  #🍆
```

will  work and is a proper tag is a bit scary and feels error prone and may make it harder to recover.

syntax/fluent.ebnf
```diff
@@ -1,41 +1,53 @@
...
+tag-list             ::= NL (__ tag)+
+
+attribute            ::= NL __ '.' identifier value
+attribute-list       ::= attribute+
+
+message              ::= identifier ((value tag-list?) | (value? attribute-list))
+value                ::= __ '=' __ pattern
```
zbraniecki: so, this opens up
```
key
  = Value
```

I have to say I don't like it from the readability perspective :(

zbraniecki: It also forbids `key=Value` format (enforcing whitespace).

syntax/fluent.ebnf
stasm: Sorry, I don't understand this. Can you provide a full example? Feel free to use `_` to denote spaces for better visibility.

syntax/fluent.ebnf
stasm: Are you asking about why `inline-space` and `break-indent` have aliases? If so, the answer is simply: readability of other productions.

syntax/fluent.ebnf
stasm: I don't understand this either. Do you mean variant keys or their patterns?

syntax/fluent.ebnf
stasm: I think we should add `#` to the list of special characters, just in case. I don't see how `🍆` is different from `żółć`, though :)

syntax/fluent.ebnf
stasm: I think @Pike wasn't convinced about the line-break before `=` either, so I'll remove it. Especially that we enforce the (pattern, tags, attrs) order anyways.

> It also forbids key=Value format (enforcing whitespace).

Would you like me to add it back?

syntax/fluent.ebnf
zbraniecki: oh, my bad, misread the syntax. Sorry!

syntax/fluent.ebnf
zbraniecki: That this enforces
```
key = { $foo ->
  *[ one ] One
}
```
whitespaces within `[`, `]`.

syntax/fluent.ebnf
zbraniecki: yeah, I just feel like i'd prefer to stay on the side of [a-z] initially and see if anyone asks for more, but maybe it'll work just fine. :)

syntax/fluent.ebnf
zbraniecki: I don't see a reason to forbid it.

syntax/fluent.ebnf
stasm: Duh, of course! Thanks, great catch :) I'll change to `_?` where needed.

syntax/fluent.ebnf
stasm: Just for Polish, I'd like to make `#męski` and `#żeński` possible, and I expect other localizers to want to use tags in their own languages, too.  This is also good for documentation and examples, as it clearly demonstrates that tags are language-specific.

PR #40, opened 2017-03-30
Change Annotation to { code, args, message, span }

stasm:
Annotations will now have unique codes: E0000 for errors, W0000 for warnings
and H0000 for hints.  They also now have a span, similar to Entries.  If an
annotation is only related to a single position in the file, the Span's start
and end field should be equal.

### DIFF
--- a/spec/fluent.asdl
+++ b/spec/fluent.asdl
@@ -31,7 +31,7 @@ module Fluent
           | Junk(str content)
           attributes (annot* annotations, span span)
 
-    annot = Annotation(str name, str message, int pos)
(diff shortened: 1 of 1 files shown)


stasm:
I made a change to the PR and added another field to `Annotation`: `args`. It's an array of string-typed arguments used to build the full message of the annotation.

This is modeled after https://docs.python.org/3/library/exceptions.html#BaseException.args and can be used to verify that annotations include extra information about parsing errors. It also opens up a way to localize annotations in CAT tools.

PR #41, opened 2017-04-11
Define the behavior of backslash escapes

stasm:
This PR is based on the comment in https://github.com/projectfluent/fluent/issues/22#issuecomment-292591531. To recap:

  - Escape sequences are only allowed in `text` and `quoted-text`.
  - Newlines are preserved by the parser. This allows proper serialization.
  - Known escape sequences are: `\\` for the literal backslash, `\"` for the literal double quote, `\{` for the literal opening brace and `\u` followed by 4 hex digits for Unicode code points. Representing code points from outside of the Basic Multilingual Plane is made possible with surrogate pairs (two `\uXXXX` sequences). Using the actual character is encouraged, however.
  - Any other escaped characters result in a parsing error. (We might relax this to producing warnings and parsing to a space for instance, but let's start with a stricter approach.)

### DIFF
--- a/spec/fluent.ebnf
+++ b/spec/fluent.ebnf
@@ -7,9 +7,7 @@ comment              ::= ('//' (char - NL)* )+
 section              ::= '[[' _? word (_ word)* _? ']]'
 
 char                 ::= [https://www.w3.org/TR/REC-xml/#NT-Char]
-
-/* special chars: { } [ \ ] */
-special              ::= [#x5b#x5c#x5d#x7b#x7d]
+hexdigit             ::= [0-9a-fA-F]
 
 /* line feed, carriage return; space, tab */
 line-break           ::= [#xA#xD]+
@@ -24,7 +22,9 @@ __                   ::= break-indent
 
 identifier           ::= [a-zA-Z_?-] [a-zA-Z0-9_?-]*
 external             ::= '$' identifier
-word                 ::= (((char - line-break) - inline-space) - special)+
+
+/* exclude whitespace and [ \ ] { } */
+word                 ::= (((char - line-break) - inline-space) - [#x5b#x5c#x5d#x7b#x7d])+
 builtin              ::= [A-Z_?-]+
 number               ::= '-'? [0-9]+ ('.' [0-9]+)?
 
@@ -43,8 +43,16 @@ attribute-list       ::= attribute+
 message              ::= identifier ((value tag-list?) | (value? attribute-list))
 value                ::= _? '=' __? pattern
 pattern              ::= (text | placeable)+
-text                 ::= ((char - line-break) - special | break-indent | '\' special)+
-quoted-text          ::= '"' (text | '\"')+ '"'
+/* text can only include newlines if they're followed by an indent */
+/* \ and { must be escaped */
+text-char            ::= (char - line-break) - [#x5c#x7b]
+                       | break-indent
+                       | '\u' hexdigit hexdigit hexdigit hexdigit
+                       | '\' [#x5c#x7b]
+text                 ::= text-char+
+/* in quoted-text " must be escaped */
+quoted-text          ::= '"' (text-char - '"' | '\"')+ '"'
+
 
 placeable            ::= '{' __? (expression | select-expression | variant-list) __? '}'
 expression           ::= quoted-text


zbraniecki:
what about multi-line, new-line first character that has special meaning? Like:
```
key = 
    This is a
    .message that I like
```

? I thought you want to backslash them as well (`.`, `#`)?

stasm:
I'd like to suggest using placeables of the form `{ "." }` and `{ "#" }` for these (rare, I presume) use-cases.  Would that work?

zbraniecki:
As a dirty hack, sure.

As a cognitive load of a "please, when you write Fluent syntax, remember to always use this dirty hack if you end up breaking a line and one of those chars end up at the beginning of the line. TY!" not so much imho.

I'd even say that I'd prefer to block multiline with arguments/tags over that.

stasm:
It looks like we disagree. The smart thing to do would be to gather some real-life user feedback and see if `\.` and `\#` are needed.

Luckily, we don't have to block 0.3 on this. Using the `{ "." }` syntax will always be a valid option. For now we can go with that as the only way to write a literal `.` and a literal `#` at the beginning of a line. `\.` and `\#` can be left as unknown escapes for now, and as such they will produce syntax errors. If we decide to add them in the future we won't have any trouble with existing translations.

What do you say?

zbraniecki:
I say that if you enable both attributes/tags and multiline you're making it backward-incompatible to ever forbid that. :)
And I say that requiring text formatting to involve context processing like "escape `.` character if it's the first character in the line" will end up being on a slide deck of someones presentation about "Bad design decisions in Fluent" one day.

stasm:
> I say that if you enable both attributes/tags and multiline you're making it backward-incompatible to ever forbid that. :)

I think I'm missing something here. Can you rephrase this? Why would we forbid that?

zbraniecki:
> I think I'm missing something here. Can you rephrase this? Why would we forbid that?

Sure. Currently, in Fluent syntax the following example may exist:

```
key =
    Share this using the hashtag:
    #projectfluent
```

Which according to your proposed semantics one meaning (one line of message value and a tag), but I claim that most users would expect it to be something else (two lines of message).

That of course is just an example, another being:


```
key = {
    [one]
        Click on this link to open another page
        [or] not.
    [two]
        Don't wait!
}
```

There are three characters - `.`, `#` and `[` that have semantic meaning (attribute, tag or variant key respectively) but can also normally be part of a string.

String can, depending on multiple factors be divided into multiple lines or not.

If you allow multiline strings and tags and attributes and variants, then you have to choose which one is recognized, and which has to be "dirty hacked".
I'm afraid that it will also be impossible to ever improve that situation once chosen.

So, I suggested that maybe for now we could, instead of establishing a dirty hack that early on in the life cycle as the only way to encode a regular character in a regular string, forbid a combination of multiline and attribute/tag/variant.

In that proposal, you could only use multiline strings as a sole value of the message, so this is allowed:

```
key = 
    <p>
        <h1>Hello World</h1>
        <span><a>Click me</a> and go on!</span>
    </p>
```

and this is allowed:

```
key = Hello World
    .label = Foo
    .accesskey = F
```

but this is not:

```
key =
    Hello World
    This is a multiline message
    .label = and this is an attribute
```

This would place the "dirty hack" instead of on a syntax, on a level of "this is not allowed" which I think is cleaner and allows us to come up with a solution to a combo of attribute/multiline that doesn't require dirty hacks, later.

Another option would be to require that multiline ends with an empty line, even if its followed by attributes/tags/variants:

```
key =
    Hello World
    This is a multiline string
    .and this is still part of it

    .label = And this is a label
```

Still not super clean, but at least no dirty hack is needed and it covers tags, labels and variants and the only cost is the required empty-line at the end of multiline string, which I think is the least invasive way to solve it.

stasm:
Thanks for the thorough explanation. It seems to be that the reason for our disagreeing is the part where you say:

> Which according to your proposed semantics one meaning (one line of message value and a tag), but I claim that most users would expect it to be something else (two lines of message).

My position is that we don't know what most users would expect here to happen.  Thus I suggest to define clear rules and don't deviate from them: a `#` at the beginning of the line defines a tag. No exceptions.

The following would parse as a value and a tag:

    key =
        Share this using the hashtag:
        #projectfluent

The following message, however, would produce an error:

    key =
        Share this using the hashtag:
        #projectfluent. Thanks!

I'd like to error as often as possible. This way we can refine the behavior farther down the road without the risk of backward-compatibility.

Introducing more rules makes it harder to follow them all at once. Forbidding tags, attrs and multiline values in a single message is one such rule that you'd have to keep in mind. I don't think I like the solution of requiring an additional blank line either. That's another rule that we introduce because we couldn't solve an issue with the syntax in a nicer way. I don't want the users to pay for that with their cognitive load.

> This would place the "dirty hack" instead of on a syntax, on a level of "this is not allowed" which I think is cleaner and allows us to come up with a solution to a combo of attribute/multiline that doesn't require dirty hacks, later.

`Foo { "bar" }` is a valid syntax and it will always be possible to write `Share this using { "#tagname" }`. Even if we come up with another solution, it will still be possible to make the edge-cases work via  `Share this using { "#tagname" }`. Let's go with this as the only way for now. I think it will suffice in general because we're talking about edge-cases, but if we see people complain about it being the only way, we can work on a better solution.

zbraniecki:
Just to be clear, I'm not blocking here. I'm merely disagreeing :)

stasm:
OK, let's merge this then and keep an eye out for feedback about the edge-cases.

PR #64, opened 2017-11-06
clarify role of `*` on selectors

hkasemir:


stasm:
Much better this way indeed. Thanks, @hkasemir!

PR #68, opened 2017-11-14
Introduce private messages.

### LINKED ISSUE #62: Introduce Glossary Messages
#### Goal

Recognize the _glossary_ translations type which cannot be retrieved by the calling code.

#### Summary

The proposed syntax for glossary messages is via an identifier starting with `-` (a dash):

```properties
-brand-name = Firefox
app-title = { -brand-name }
```

Glossary messages cannot be retrieved from the `MessageContext` they're defined in and they can only be referenced in other messages. Glossary messages and public messages are separate AST nodes extending the `message` node in the ASDL spec.

#### Description

Glossary translations are not intended to be used directly by the calling code. Instead they should go through another, public, message.  The current syntax already hints at this by disallowing tags and attributes to be defined on the same message. The introduction of glossary messages will make this design explicit.

```properties
-brand-name = Firefox
application-title = { -brand-name }
```

This change boils down to allowing `-` as the first character of the identifier. Its consequences are, however, far-reaching. Tools like `compare-locales` should not inspect attributes of glossary messages. These attributes are consequently private and cannot be used for translating widget attributes (like in HTML). They may be used for language-specific information and they could replace tags.

#### Example

English:

```properties
-brand-name = Firefox

app-title = { -brand-name }
has-updated = { -brand-name } has been updated.
```

Polish (with glossary attributes replacing tags):

```properties
-brand-name =
    {
       *[nominative] Firefox
        [genitive] Firefoksa
    }
    .gender = masculine

app-title = { -brand-name }
has-updated =
    { -brand-name.gender ->
        [masculine] { -brand-name } został zaktualizowany.
        [feminine] { -brand-name } została zaktualizowana.
       *[other] Program { -brand-name } został zaktualizowany.
    }
```

#### Private Messages

By allowing `-` as the first character of the identifier we will also enable ids of the form `--foo`.  These could be used by tools to differentiate between glossary messages which must be present in an FTL file and _private_ messages which are localizer-defined.

```properties
--tab = tab
open-tab = Open a new { --tab }
```

I would still encourage everyone to spell the whole sentence out:

```properties
open-tab = Open a new tab
```

#### Signoffs

  - [x] @Pike 
  - [x] @stasm
  - [x] @zbraniecki

stasm:
Fix #62.

Private messages are similar to regular messages but they can only be
used as references in other messages. Their identifiers start with at
least one dash `-`, e.g. `-brand-name`. The runtime cannot retrieve
private messages directly. They are best used to define glossary terms
which can be used consistently across the localization of the entire
product.

Private messages can store some grammatical information about themselves
in tags. Private messages are not allowed to have attributes.

### DIFF
--- a/guide/external.md
+++ b/guide/external.md
@@ -1,5 +1,13 @@
 # Interpolation and External Arguments
 
+Strings in FTL may use special syntax to incorporate small pieces of
+programmable interface. Those pieces are denoted with curly braces `{` and `}`
+and are called placeables.
+
+It's common to use placeables to interpolate external arguments into the
+translation. External arguments are provided by the developer and may change
+on runtime.
+
 ```
 welcome = Welcome { $user }
 unreadEmails = { $user } has { $emailCount } unread emails.
@@ -10,9 +18,6 @@ unreadEmails = { $user } has { $emailCount } unread emails.
     "emailCount": 5
 }
 ```
-
-Another common use case for a placeable is to put an external argument,
-provided by the developer, into the string.
 
 There are all kinds of external data that might be useful in providing a good
 localization: user names, number of unread messages, battery level, current
--- a/guide/private.md
+++ b/guide/private.md
@@ -1,0 +1,19 @@
+# Private Messages
+
+```
+-brand-name = Firefox
+
+app-title = { -brand-name }
+has-updated = { -brand-name } has been updated.
+```
+
+Private messages are similar to regular messages but they can only be used as
+references in other messages. Their identifiers start with at least one dash
+`-` like in the example above: `-brand-name`. The runtime cannot retrieve
+private messages directly. They are best used to define terms which can be
+used consistently across the localization of the entire product.
+
+Private messages can store some grammatical information about themselves in
+[tags](tags.html) which are the topic of the next chapter.
+
+Private messages are not allowed to have [attributes](attributes.html).
--- a/spec/fluent.asdl
+++ b/spec/fluent.asdl
@@ -25,11 +25,15 @@ module Fluent
 {
     res = Resource(entry* body, comment? comment, span? span)
 
-    entry = Message(iden id, pat? value, attr* attributes, tag* tags, comment? comment)
+    entry = BaseMessage(base_message)
           | Section(symb name, comment? comment)
           | Comment(comment)
           | Junk(str content)
           attributes (annot* annotations, span? span)
+
+    base_message = Message(pat? value, attr* attributes)
+                 | PrivateMessage(pat value, tag* tags)
+                 attributes (iden id, comment? comment)
 
     annot = Annotation(str code, str* args, str message, span? span)
     span = Span(int start, int end)
--- a/spec/fluent.asdl
+++ b/spec/fluent.asdl
@@ -44,19 +44,16 @@ module Fluent
          | Placeable(expr expression, span? span)
 
     -- Expressions
-    expr = InlineExpression(inline_expr)
-         | BlockExpression(block_expr)
-    inline_expr = StringExpression(string)
-                | NumberExpression(number)
-                | MessageReference(iden id)
-                | ExternalArgument(iden id)
-                | AttributeExpression(iden id, iden name)
-                | VariantExpression(iden id, varkey key)
-                | CallExpression(fun callee, arg* arguments)
-                | Placeable(expr expression)
-                attributes (span? span)
-    block_expr = SelectExpression(inline_expr? expression, var* variants)
-               attributes (span? span)
+    expr = StringExpression(string)
+           | NumberExpression(number)
+           | MessageReference(iden id)
+           | ExternalArgument(iden id)
+           | AttributeExpression(iden id, iden name)
+           | VariantExpression(iden id, varkey key)
+           | CallExpression(fun callee, arg* arguments)
+           | SelectExpression(expr? expression, var* variants)
+           | Placeable(expr expression)
+           attributes (span? span)
 
     -- Attributes of Message
     attr = Attribute(iden id, pat value, span? span)
@@ -70,7 +67,7 @@ module Fluent
            | Symbol(symb)
 
     -- Arguments to CallExpression
-    arg = Expression(inline_expr)
+    arg = Expression(expr)
         | NamedArgument(iden name, argval value, span? span)
     argval = NumberExpression(number)
            | StringExpression(string)
--- a/spec/fluent.ebnf
+++ b/spec/fluent.ebnf
@@ -22,7 +22,9 @@ NL                   ::= line-break
 __                   ::= break-indent
 
 identifier           ::= [a-zA-Z] [a-zA-Z0-9_-]*
-external             ::= '$' identifier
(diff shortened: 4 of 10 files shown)


spec/fluent.asdl
```diff
@@ -25,12 +25,16 @@ module Fluent
...
 
-    entry = Message(iden id, pat? value, attr* attributes, tag* tags, comment? comment)
+    entry = BaseMessage(base_message)
           | Section(symb name, comment? comment)
           | Comment(comment)
           | Junk(str content)
           attributes (annot* annotations, span? span)
 
+    base_message = Message(pat? value, attr* attributes)
+                 | PrivateMessage(pat value, tag* tags)
```
stasm: This makes it a requirement that `PrivateMessages` have a non-null `value`. Should we enforce this for now or keep it consistent with public `Messages`?

zbraniecki: I'm ok enforcing it.

spec/fluent.ebnf
```diff
@@ -21,7 +21,9 @@ _                    ::= inline-space
 NL                   ::= line-break
 __                   ::= break-indent
 
-identifier           ::= [a-zA-Z] [a-zA-Z0-9_-]*
+public-identifier    ::= [a-zA-Z] [a-zA-Z0-9_-]*
+private-identifier   ::= '-'+ public-identifier
+identifier           ::= public-identifier | private-identifier
 external             ::= '$' identifier
```
zbraniecki: I think you need to change `external` to only accept `public-identifier`. Or, maybe I'd even change `public-identifier` to `identifier`, and then create `message-identifier ::= identifier | private-identifier`.

spec/fluent.ebnf
```diff
@@ -41,7 +43,9 @@ tag-list             ::= NL (__ tag)+
 attribute            ::= NL __ '.' identifier value
```
zbraniecki: same here.

spec/fluent.ebnf
stasm: Good call, I like this approach better. Thanks.

spec/fluent.ebnf
```diff
@@ -69,8 +73,8 @@ block-expression     ::= select-expression
                        | variant-list
 
 select-expression    ::= inline-expression __ '->' __ variant-list
-attribute-expression ::= identifier '.' identifier
-variant-expression   ::= identifier '[' _? variant-key _? ']'
+attribute-expression ::= message-identifier '.' identifier
```
Pike: If this is just `identifier`, the syntax defines that we can't dereference attributes of private messages?

That'd make inline-expression safe, and we could use an unsafe-inline-expression that adds private-indentifier.identifier as selector-expression instead of inline-expression?

stasm: Nice, I like this approach. It makes syntax work for us without adding more... syntax.

We could even go further and define `selector-expression` which would be different from `inline-expression` and `block-expression` and could comprise of:

```
inline-expression    ::= quoted-text
                       | number
                       | identifier
                       | external
                       | variant-expression
                       | call-expression
                       | placeable

block-expression     ::= select-expression
                       | variant-list

selector-expression  ::= quoted-text
                       | number
                       | identifier
                       | external
                       | attribute-expression
                       | call-expression
```

Compare it with:

https://github.com/projectfluent/fluent/blob/b9552ae754c373311e79594a73a90a29b4540ef5/spec/fluent.ebnf#L60

Pike: Yeah, agreed, allowing placeable or variant-expression doesn't make much sense in a selector-expression

stasm:
> The asdl changes are a bit above my head, mostly because I don't know if there's a difference between a param thingie and an attributes thingie.

`attributes` can be used to avoid repetition of fields in unions. The following two notations are equivalent:

    prod = Foo(string name)
         | Bar(string name)

    prod = Foo
         | Bar
         attributes (string name)

The relevant excerpt from [The Zephyr Abstract Syntax Description Language](ftp://ftp.cs.princeton.edu/techreports/1997/554.pdf), p. 5, § 2.6:

> Often  several  constructors  of  a  type  share  a  set  of common values.  To make this explicit, ASDL includes an attribute notation.  Since all variants of a type carry the values of an attribute, its fields can be accessed without having to discriminate between the various constructors.  Attributes can be seen as providing some limited features of inheritance

guide/external.md
```diff
@@ -1,5 +1,12 @@
 # Interpolation and External Arguments
 
+Strings in FTL may use special syntax to incorporate small pieces of
+programmable interface. Those pieces are denoted with curly braces `{` and `}`
+and are called placeables.
+
+It's common to used placeables to interpolate an external argument, provided
+by the developer, into the string.
```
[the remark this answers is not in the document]
stasm: `It's common to use placeables to interpolate an external argument into the string, provided by the developer` doesn't sound right either. I'll go with `It's common to use placeables to interpolate external arguments into the translation. External arguments are provided by the developer and may change on runtime.`

guide/private.md
```diff
@@ -0,0 +1,19 @@
...
+-brand-name = Firefox
+
+app-title = { -brand-name }
+has-updated = { -brand-name } has been updated.
+```
+
+Private messages are similar to regular messages but they can only be used as
+references in other messages. Their identifiers start with at least one dash
+`-` like in the example above: `-brand-name`. The runtime cannot retrieve
+private messages directly. They are best used to define glossary terms which
```
Pike: Just dropping the `glossary` here alltogether sounds good enough for me.

spec/fluent.asdl
```diff
@@ -40,19 +44,16 @@ module Fluent
          | Placeable(expr expression, span? span)
 
     -- Expressions
-    expr = InlineExpression(inline_expr)
-         | BlockExpression(block_expr)
-    inline_expr = StringExpression(string)
```
Pike: `inline_expr` is still used in line 70?

    arg = Expression(inline_expr)

spec/fluent.ebnf
```diff
@@ -54,24 +57,30 @@ text                 ::= text-char+
...
                        | variant-list
 
-select-expression    ::= inline-expression __ '->' __ variant-list
-attribute-expression ::= identifier '.' identifier
-variant-expression   ::= identifier '[' _? variant-key _? ']'
+select-expression    ::= selector-expression __ '->' __ variant-list
+selector-expression  ::= quoted-text
+                       | number
+                       | external-identifier
+                       | public-identifier
```
Pike: This should be private-identifier instead of public-identifier, as tags are only on private messages, and I guess the intent is to not select based on translation value?

stasm: Good catch and yes, that's the intent. At least for now.

spec/fluent.asdl
stasm: Thanks.

guide/private.md
stasm: Just to make sure I understand this comment, you'd rather see `…to define terms which…`?

Pike: yes.

PR #69, opened 2017-11-15
 Replace tags with attributes defined on private messages. 

### LINKED ISSUE #67: Remove tags
This is a counter-proposal to #59. If #62 is approved we could go further than only allowing tags on glossary/private messages and replace tags with attributes defined on private messages.

The main reason why tags exist is to encode language-specific data like grammatical genders etc. We decided to add a separate data structure so that it's easy to distinguish between data which is part of the message's public interface and data which is private. In other words, the use-case for tags is tooling.

With private/glossary messages we satisfy this use-case by marking the message as special. Tools can choose to check for existence of glossary messages but ignore their attributes. In such scenario, those attributes can be used to store language-specific information about the translation.

By removing tags we:

  - remove an entire data type which makes Fluent easier to understand,
  - free up the tag's sigil making the syntax easier to learn,
  - simplify the message lookup in selectors.

Wrt. the last point, right now tag lookup simply uses the message reference in the selector. Messages are not matched to variant keys by value but rather by their tags. In case more than one tag are present, any matching tag will select the variant. We haven't yet defined a way to AND tags in variant keys.

```properties
brand-name = Aurora
    #żeński

has-updated =
    { brand-name ->
        [męski] { brand-name} został zaktualizowany.
        [żeński] { brand-name } została zaktualizowana.
       *[inny] Program { brand-name } został zaktualizowany.
    }
```

Without tags, the above becomes a selector with an `AttributeExpression` on a private message:

```properties
-brand-name = Aurora
    .gender = żeński

has-updated =
    { -brand-name.gender ->
        [męski] { -brand-name} został zaktualizowany.
        [żeński] { -brand-name } została zaktualizowana.
       *[inny] Program { -brand-name } został zaktualizowany.
    }
```

Once we implement #4 it will become easy and consistent with the rest of the syntax to use more than one private attribute for the selection logic.

#### Signoffs

  - [x] @Pike 
  - [x] @stasm
  - [x] @zbraniecki

stasm:
Fixes #67. Requires #68 (included in this PR). Please only review the last commit.


spec/fluent.asdl
```diff
@@ -31,9 +31,9 @@ module Fluent
           | Junk(str content)
           attributes (annot* annotations, span? span)
 
-    base_message = Message(pat? value, attr* attributes)
-                 | PrivateMessage(pat value, tag* tags)
-                 attributes (iden id, comment? comment)
+    base_message = Message
+                 | PrivateMessage
+                 attributes (iden id, pat? value, attr* attributes, comment? comment)
```
stasm: #68 made `value` required on `PrivateMessage`. That's because a message consisting only of tags doesn't make much sense. With tags removed, I still think it's unlikely to have private messages consisting only of attributes. Allowing it, however, makes private messages consistent with public messages. It reduces the number of special rules to remember.

Pike: I'd prefer to keep the requirement for private messages to have a value, otherwise programmers will abuse the uncontrollable hashmap of a private public attribute thing.

guide/private.md
```diff
@@ -13,7 +13,27 @@ references in other messages. Their identifiers start with at least one dash
...
+        [feminine] { -brand-name } została zaktualizowana.
+       *[other] Program { -brand-name } został zaktualizowany.
+    }
+```
+
+Use attributes to describe grammatical traits and properties. Genders, animacy, whether the private message starts with a vowel or not etc.
+
+Use variants to define different _facets_ of the message. Variants are
+essentially the same value, just in slightly different forms to make it
+grammatically correct when used inside of other messages.
```
Pike: Add a paragraph that attributes of private message are not publicly accessible?

spec/fluent.asdl
stasm: Fair point. We can enforce it in the EBNF and keep the ASDL simple.

stasm:
> We need to change the private-identifier to private-attribute-expression in selector-expression.

I'm somewhat confused because I remember making this change after I merged #68. Perhaps I lost in the rebase that followed or didn't save it :/ Thanks for catching this; juggling these PRs is not exactly easy.

> I haven't looked at the docs outside of the patch, I guess it'd be good to have another pass on them before we release 0.5

Yes, it would be good to take a broader look and I plan on doing it. I tried to make sure they make at least some sense between those PRs.

> And I've kinda given up on the asdl. Not sure what it's use-case is right now, we should either fix it of fix it (via git rm ;-) )

While the EBNF describes the input of the parser, the ASDL describes its output. It's a cross-language description of:

  - https://github.com/projectfluent/fluent.js/blob/master/fluent-syntax/src/ast.js,
  - https://github.com/projectfluent/python-fluent/blob/master/fluent/syntax/ast.py, and
  - https://github.com/projectfluent/fluent-rs/blob/master/src/syntax/ast.rs.

It's the closest to `ast.rs` because ASDL has strict typing. As someone who works with all three parsers, I find the ASDL very useful. It's also something tooling authors can refer to for a bird-eye overview of the AST.

PR #70, opened 2017-11-15
Comment levels

### LINKED ISSUE #58: Comment Levels
#### Goal

Allow comments to define the outline of the FTL file in a Markdown-like fashion.

#### Description

This proposal is inspired by Markdown and how multiple `#`, `##`, `###` etc. can be used to define headers which in turn define the outline of the document.

In FTL, the number of `#` would correspond to the comment level:

  - `#` stands for a regular comment: standalone or attached to a message,
  - `##` stands for a standalone level-2 comment, or a group-level comment,
  - `###` stands for a standalone level-3 comment, or a resource-level comment.

FTL currently allows sections (`[[ Section Name ]]`) for defining the outline of the FTL file. Entries under a section are not explicitly grouped in the AST but tools may use sections for grouping messages.  Looking at some [examples of sections in the wild](https://github.com/mozilla/testpilot/blob/99573605ac85bd34c40efc1740de595bfccceb90/locales/en-US/app.ftl) it's clear that sections are mostly used for their comments. I propose to remove sections and replace them with group-level comments (`##`) to streamline this practice.

#### Example

```properties
# This Source Code Form is subject to the terms of the Mozilla Public
# License, v. 2.0. If a copy of the MPL was not distributed with this
# file, You can obtain one at http://mozilla.org/MPL/2.0/.

### Localization for Server-side strings of Firefox Screenshots
### Please don't localize Firefox, Firefox Screenshots, or Screenshots

## Global phrases shared across pages

my-shots = My Shots
home-link = Home
screenshots-description =
    Screenshots made simple. Take, save, and
    share screenshots without leaving Firefox.

## Creating page

# Note: { $title } is a placeholder for the title of the web page
# captured in the screenshot. The default, for pages without titles, is
# creating-page-title-default.
creating-page-title = Creating { $title }
creating-page-title-default = page
creating-page-wait-message = Saving your shot…
```

#### Details

A change to the comment level automatically starts a new comment. The above example could also be written without the blank line after `## Creating page`:

```properties
## Creating page
# Note: { $title } is a placeholder for the title of the web page
# captured in the screenshot. The default, for pages without titles, is
# creating-page-title-default.
creating-page-title = Creating { $title }
creating-page-title-default = page
creating-page-wait-message = Saving your shot…
```

Only `#` comments can be attached to a message. In the above example we can also remove the blank line after `## Global phrases shared across pages`:

```properties
## Global phrases shared across pages
my-shots = My Shots
home-link = Home
screenshots-description =
    Screenshots made simple. Take, save, and
    share screenshots without leaving Firefox.
```

Because `###` would now uniquely identify resource-level comments, I would also like to remove the `comment` field from the `Resource` production and remove the parser heuristic that treats the first standalone comment as the resource comment. 

Commenting out parts of the FTL file is still possible although it should whitespace to make it unambiguous: 

```properties
# ## Creating page
# # Note: { $title } is a placeholder for the title of the web page
# # captured in the screenshot. The default, for pages without titles, is
# # creating-page-title-default.
# creating-page-title = Creating { $title }
# creating-page-title-default = page
# creating-page-wait-message = Saving your shot…
```

#### Dependencies

If this proposal gets enough support, I'd like to reclaim `#` for comments and use a different sigil for tags.

We should also plan a transition period during which both the comment level syntax and the current section syntax is still valid. This will make it much easier to handle migrations in live projects.

stasm:
Fixes #58. Requires #68 and #69 (both of which are included in this PR right now). Please only review the last commit.


spec/fluent.asdl
```diff
@@ -73,10 +72,10 @@ module Fluent
...
     
     iden = Identifier(str name, span? span)
-    symb = Symbol(str name, span? span)
+    varname = VariantName(str name, span? span)
     fun = Function(str name, span? span)
     string = StringExpression(str value, span? span)
     number = NumberExpression(str value, span? span)
 
-    comment = Comment(str content, span? span)
+    comment = Comment(int level, str content, span? span)
```
stasm: Would it better to introduce three node types instead of the `level` field? E.g. `SingleComment`, `DoubleComment` and `TripleComment`.

spec/fluent.asdl
```diff
@@ -23,10 +23,9 @@
 
 module Fluent
 {
-    res = Resource(entry* body, comment? comment, span? span)
+    res = Resource(entry* body, span? span)
```
zbraniecki: What is the expected outcome of "file-level" comment being placed in the middle of the file? Say:

```properties
key = Message

# Message level comment
key2 = Message 2

### Suddenly, a wild file-level message appears
key3 = Message 3
```

spec/fluent.asdl
zbraniecki: Not sure which way to go. The concern I have is about semantically correct mixed level comments:

```properties
# Foo
### Bar
# Baz
## Qux
key = Value
```

stasm: These would be four independent `Comment` entries with levels: `1`, `3`, `1`, and `2`. Or four independent entries: `SingleComment`, `TripleComment`, `SingleComment`, and a `DoubleComment`.

spec/fluent.asdl
stasm: It appears in the AST a such: a `Comment` entry between `key2` and `key3`.

stasm:
> Before this PR it's very clear to me that a comment at the very top of the file is "attached" to it and thus is a "File Level Comment".

FWIW, this has never been as clear to me. I find it confusing that the the file-level comment is the first inside of the file while the message-level comment is the first outside (and before) the message.

> A comment attached to a message is a "Message Level Comment" and so on.

This doesn't change for `#` comments. Only `#` comments can be attached to a message.

> With this changes I feel like we're getting into a world where there may be multiple "File Level Comments" spread around the file (…)

Yes, that's now possible, but I expect the linter to warn against this. In most cases we would only care about the first file-level comment.

> the distinction between a comment attached to a Message and "standalone" is also unclear to me.
What should happen if the comment attached to a Message is a Group Level Comment? Or File Level Comment?

Only `#` comments can be attached to a message. Thus the following:

```properties
## Group Level Comment
foo = Foo
```

…parses as two separate entries: a `Comment` and a `Message`.

stasm:
> With the new system, I don't see a good way except of some enforced best-practice to make sure that the first line of the group level comment actually is a "title" of such group and the following lines are describing the group.

With this change I'm betting on the fact the developers don't care about the titles; they care about the comment. For example ([source](https://github.com/mozilla/testpilot/blob/0bac5a2c7213f07f818aefcf5464cd01b2ef2ebc/locales/en-US/app.ftl#L3-L8)):

```
// Page titles, put in the <title> HTML tag.
[[pageTitle]]
pageTitleDefault = Firefox Test Pilot
pageTitleLandingPage = Firefox Test Pilot
pageTitleExperimentListPage = Firefox Test Pilot - Experiments
pageTitleExperiment = Firefox Test Pilot - {$title}
```

i think Pontoon should show the entire Group Comment for each message.

spec/CHANGELOG.md
```diff
@@ -28,6 +28,18 @@
...
+  - Removed sections and introduced comment levels. (#58)
+
+    Comments starting with `##` and `###` define the outline of the file and
+    replace sections. `##` denotes group-level comments and `###` denotes
+    file-level comments. These are always standalone `Comment` entries. The
+    `Comment` node has gained the `level` field. The `comment` field on the
+    `Resource` node has been removed.
+
+  - Renamed `Symbol` to `VariantName` in the ASDL.
+
```
Pike: Should this add something about backwards compatiblity? Is this something we do on the fluent side or on the implementations side?

If it's impl, should there be something here that suggests that impls do this?

spec/fluent.ebnf
```diff
@@ -31,8 +32,7 @@ word                 ::= (((char - line-break) - inline-space) - [#x5b#x5c#x5d#x
 builtin              ::= [A-Z] [A-Z_?-]*
 number               ::= '-'? [0-9]+ ('.' [0-9]+)?
 
-variant-key          ::= number | variant-symbol
-variant-symbol       ::= word (_ word)*
+variant-key          ::= word (_ word)*
```
Pike: Is this change intentional?

stasm: Yes. The `word` production already allows all characters that `number` did.

stasm: Hmm, but having two numbers next to each other separated by a space is not what we want. OK, I'll bring this back.

spec/CHANGELOG.md
stasm: What do you mean by _add something_?

Pike: Should we document how we expect implementations to deal with the update from syntax 0.4 to 0.5?

stasm: I think we should suggest a way forward when we release 0.5. It will be based on the approach we'll take with our implementations.

spec/fluent.asdl
stasm: > Would it better to introduce three node types instead of the level field? E.g. `SingleComment`, `DoubleComment` and `TripleComment`.

@Pike, do you have an opinion or a recommendation about this?

Pike: I wouldn't call 'em by markup, but by purpose. So, `FileComment`, `GroupComment`, `MessageComment` or just `Comment`. Or just leave them with levels.

stasm: I went for `Comment`, `GroupComment` and `FileComment`. They are explicit and clearly communicate the role. They also define the interpretation of each of the comment types right at the level of the syntax which will will improve the portability of FTL files.

guide/comments.md
```diff
@@ -1,29 +1,46 @@
...
+
+```properties
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+### Localization for Server-side strings of Firefox Screenshots
+### Please don't localize Firefox, Firefox Screenshots, or Screenshots
+
+## Global phrases shared across pages
```
zbraniecki:
```
## Global phrases
##
## Phrases shared accross pages.
```

spec/fluent.ebnf
```diff
@@ -3,8 +3,9 @@ entry                ::= comment
                        | section
                        | message
 
-comment              ::= ('//' (char - NL)* NL)+
-section              ::= '[[' _? word (_ word)* _? ']]'
+comment              ::= ('# ' (char - NL)* NL)+
+                       | ('## ' (char - NL)* NL)+
+                       | ('### ' (char - NL)* NL)+
```
zbraniecki: switch to `inline-space` instead of ` space char following the sygil

guide/comments.md
Pike: Not sure that's readable in tools like pontoon. I think we should advertise single-line or at least format-flowed comments.

guide/comments.md
```diff
@@ -1,29 +1,46 @@
...
+a comment is located right above a message it is considered part of the
+message and localization tools will present the message and the comment
+together. Otherwise the comment is standalone (which is useful for commenting
+parts of the file out).
+
+Double-hash comments (`##`) are always standalone. They can be used to divide
+files into smaller groups of messages related to each other; they are
+group-level comments. Think of them as of headers with a description.
+Group-level comments are intended as a hint for localizers and tools about
+the layout of the localization resource.
```
Pike: Can you add docs on how to end a group? Aka, something like

    The grouping ends with the next group comment. Use an empty group comment if there's no following group.

Or something not that butchered.

guide/comments.md
stasm: If I understand @zbraniecki's intent correctly, Pontoon could display only the first line of the comment in places where screen space is tight. This would be similar to how Git treats commit messages.

This is a good practice that we should discuss separately. For now, I made all comments in the PR  single-line.

PR #74, opened 2017-12-06
Syntax sugar for block expression values

### LINKED ISSUE #42: Allowing full-message variants to end without indention
While working with Fluent syntax, I'm consistently hitting this awkward moment when this:

```
key = {
    $foo
}
```

is a parser error, while this:

```
key = {
        $foo
    }
```

is the way to go.

I believe we should allow for the former.

stasm:
Fix #42.

### DIFF
--- a/spec/fluent.ebnf
+++ b/spec/fluent.ebnf
@@ -1,11 +1,11 @@
-body                 ::= blank-line* (entry NL blank-line*)* entry? EOF
+body                 ::= blank-line* (entry line-break blank-line*)* entry? EOF
 entry                ::= comment
                        | section
                        | message
 
-comment              ::= ('#' (inline-space (char - NL)*)? NL)+
-                       | ('##' (inline-space (char - NL)*)? NL)+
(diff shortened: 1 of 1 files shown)


spec/fluent.ebnf
```diff
@@ -32,17 +32,17 @@ number               ::= '-'? [0-9]+ ('.' [0-9]+)?
...
 
-attribute            ::= NL __ '.' public-identifier value
+attribute            ::= __ '.' public-identifier value
 attribute-list       ::= attribute+
 
 message              ::= public-identifier ((value attribute-list?) | (value? attribute-list))
                        | private-identifier value attribute-list?
 
-value                ::= _? '=' __? pattern
+value                ::= _? '=' __? (pattern | '{' __? block-expression blank-line* '}')
```
zbraniecki: I'm wondering if it should be:

```
value                ::= _? '=' (__? pattern | _? '{' __? block-expression blank-line* '}')
```

stasm: Yes, good idea, thanks.

PR #76, opened 2017-12-14
Streamline the ASDL for Messages

stasm:
We can rely on checking the first character of the identifier to detect private messages.

In the runtime, we only need this check to forbid retrieving private messages via `MessageContext.getMessage`. All other use-cases are eagerly handled in the parsing and result in syntax errors.


Pike:
Cancelling the review request on me, I don't have much opinions on the asdl.

PR #83, opened 2018-01-25
Specify bounds of quoted text

stasm:
Let's explicitly allow zero-width `quoted-text` and forbid it to span multiple lines.

### DIFF
--- a/spec/fluent.ebnf
+++ b/spec/fluent.ebnf
@@ -43,13 +43,13 @@ message              ::= identifier (value attribute* | attribute+)
 
 value                ::= _? '=' __? pattern
 pattern              ::= (text | placeable)+
-/* text can only include newlines if they're followed by an indent */
+
 /* \ and { must be escaped */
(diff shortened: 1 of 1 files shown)


spec/fluent.ebnf
```diff
@@ -43,15 +43,15 @@ message              ::= identifier (value attribute* | attribute+)
 
 value                ::= _? '=' __? pattern
 pattern              ::= (text | placeable)+
-/* text can only include newlines if they're followed by an indent */
+
```
Pike: meh, cannot comment on the line below, but ... no ` and `

stasm: Do you mean no `and` in `\ and { must be escaped`? `\` and `{` correspond to `#x5c#x7b` which this rule says must be escaped.

Pike: oh goodness, yeah, that comment is really hard to read. Sorry. `\` just slipped my mind here.

Issue #102, opened 2018-04-27
New value type: Variant List


stasm:
We currently allow Select Expressions to have a null selector. This is mostly useful in Terms to define facets of the term's value:

```properties
-brand-name =
    {
       *[nominative] Firefox
        [genitive] Firefox's
    }
```

The grammar defines the above syntax as:

```yml
Term
    id: Identifier
        name: "-brand-name"
    value: Pattern
        elements:
            - Placeable
                expression: SelectExpression
                    expression: null
                    variants: […]
```
Variants defined this way can be retrieved inside of Placeable with the `-term[variant name]` syntax.  The runtime checks if the term's value is a pattern with a single element: a placeable which is a selector-less select expression. If the conditions are met, the variant is retrieved successfully.

The fact that we need this complex runtime check indicates to me that we're abusing Patterns for a logic-less (no selector) store for variants.

I'd like to make the design of Variant List more explicit by making them a new value type. Grammar-wise, values of Messages, Terms and Variants could either be Patterns or Variant Lists. Value List would not be expressions. Static analysis tools can then very easily tell when Variant Lists are mis-used: they shouldn't appear in Message at all, and they should only be nested inside of other Variant Lists.

The FTL snippet above would parse as follows:

```yml
Term
    id: Identifier
        name: "-brand-name"
    value: VariantList
        variants: […]
```

Pike:
There are aspects here that are not about if we do it or not, but where. Is it Syntax or Semantics that are statically checked.

I'd like to see both variants (pun intended), and see which of the two is easier to learn. We'll need to document both, one will be solely in the syntax, the other will be split between syntax and semantics.

zbraniecki:
I like this proposal. I see a potential in the VariantList in the future so I'm happy to see it being formalized.

stasm:
I think we talked about forbidding `VariantList` in `Messages` all together. I implemented this PR based on the [Syntax Semantics](https://github.com/projectfluent/fluent/wiki/Syntax-Semantics) wiki page, but it didn't have any rules for `Messages` drafted. Thanks for the reminder. I now edited it with a proposal which I hope reflects what we talked about. I'll work on implementing these rules in this PR on Monday and I'll re-request review. It would help to land #128 first to avoid conflicts.

stasm:
[...]

Issue #99, opened 2018-04-27
Define and document linting rules for Fluent


Pike:
We talk a lot about linters these days, but we don't have a place where we document Fluent's intent for that.

I think we should also find a name for that doc that isn't *Linting*, as I consider that to be an implementation detail. I talked about *Well-formed* vs *Valid* in #97, maybe that's a good start?

Where should we put this? projectfluent.org or on the wiki here?

CC @stasm, @flodolo, @hkasemir, @zbraniecki

stasm:
We definitely needs this. Currently way too often do we use _linting_ as the silver bullet answer for all syntax questions :)

Let's use the wiki first so that we can edit freely. I like the distinction between _well-formed_ and _valid_. In the XML meaning of the words, would _validation_ require some kind of additional metadata about the available builtins?

zbraniecki:
I like rust's separation between `cargo check`, `cargo fmt` and `cargo clippy` - check verifies that the syntax is correct, fmt normalizes the formatting, clippy advises on the use of the language.

I imagine our `check` warning on missing variables, plural categories and report erros, fmt normalizing whitespace, while clippy advising on use of redundant parts of a message, or potential refactors.

zbraniecki:
is there actually a use case which wants `did it parse?` but not `does it make sense?`? In rust case, `undefined.prop` would raise a warning in `cargo check`

stasm:
> is there actually a use case which wants did it parse? but not does it make sense??

Yes, it's editing FTL. Making it easier to write well-formed FTL means less frustration because the thing you just wrote didn't parse at all. In fact, by making it parse, you can easily syntax-highlight it _and_ run static analysis to give hints and warnings about the semantics.

Rust is infamous for not being very friendly at first and many people describe the experience as _wrestling_ with the compiler. I think it's OK for Rust to require some effort to learn. I don't think, however, that localizers have the same budget of patience as aspiring Rust developers. I wouldn't want anyone to _wrestle_ with Fluent :)

> In rust case, undefined.prop would raise a warning in cargo check

Right, but there are two separate steps involved. First, the syntax is parsed and it produces a valid AST. Then, `rustc` imbues this AST with semantics—and it warns about something which makes little sense.

stasm:
I created a stub of validation rules at https://github.com/projectfluent/fluent/wiki/Syntax-Semantics. I'm happy to extend it when I'm back from vacation next week.

[...]

PR #103, opened 2018-05-15
Implement the reference Fluent Syntax parser

stasm:
The _kombi_ parser started out as an experiment with LL parser combinators two weeks ago when I met with @Pike in Berlin. The initial prototype lived in https://github.com/stasm/fluent-kombi. I refined it to a point where it's ready to become the reference parser implementation.

This PR removes `diagram.xhtml` because the EBNF generated from `syntax/grammar.mjs` contains RegExps, which are not parsed by the [Railroad Diagrams Generator](http://www.bottlecaps.de/rr/ui). I expect we'll use more RegExps in the future for performance reasons, but for now I deliberately only used it in a few places (mostly as `/.*/`). The PR also replaces the ASDL with `syntax/ast.mjs` defining the AST data structures and `syntax/abstract.mjs` defining how to transforms the CST (concrete parse tree, the direct output of the parser) into the AST.

Unfortunately, I wasn't yet able to use the [fixtures from `fluent-syntax`](https://github.com/projectfluent/fluent.js/tree/master/fluent-syntax/test) because the reference parser doesn't produce any annotations and doesn't add spans to AST nodes. I'd like to add both of these features in the future. I added a new set of fixtures and in some cases I copied the FTL fixtures from `fluent-syntax` and generated new JSON AST for them. The current coverage of the tests is similar to the `fixtures_structure` one's from `fluent-syntax`.


syntax/grammar.mjs
```diff
@@ -0,0 +1,495 @@
...
+        between(
+            maybe(inline_space),
+            char("=")),
+        Pattern.as("value"),
+        repeat(Attribute).as("attributes"),
+        line_end)
+    .map(to_object)
+    .map(into(FTL.Term)));
+
+let Attribute = defer(() =>
```
Pike: How about moving that until after the pattern block? I find it's position here to be a bit of a break in reading the grammar and the ebnf.

syntax/grammar.mjs
```diff
@@ -0,0 +1,495 @@
...
+        sequence(
+            string("###"),
+            comment_line.as("line")))
+    .map(flatten(1))
+    .map(to_array)
+    .map(to_string)
+    .map(into(FTL.ResourceComment)));
+
+let Junk = defer(() =>
+    repeat1(
```
Pike: Looking at how this renders and reads, what do you think about making Junk just single line in the syntax, and glue Junk nodes together in in abstract.mjs?

That would Junk just make "eat the rest of the line", right?

syntax/grammar.mjs
```diff
@@ -0,0 +1,495 @@
...
+
+let ExternalArgument = defer(() =>
+    ExternalIdentifier.map(into(FTL.ExternalArgument)));
+
+let CallExpression = defer(() =>
+    sequence(
+        Function.as("callee"),
+        char("("),
+        between(
+            maybe(inline_space),
```
Pike: I think this reads better without the `between inline_space`. We have that around each argument anyway.

stasm: We need at least the `maybe(inline_space)` part to handle the case of no arguments: `FUNCTION( <whitespace here>     )`.

syntax/grammar.mjs
stasm: Good point. I'll move it into it's own comment group right above `Pattern`, because `Attribute` references it as well and I'd like to keep all `Expressions` right below `Pattern`.

syntax/grammar.mjs
stasm: Interesting idea, that would work, yes. It would only avoid the `repeat1` though, right? I think I'd prefer to keep it as it is right now. In fact, I already have a few ideas for changes that will reduce the amount of post-processing in `abstract.mjs`. If something can be expressed in the grammar, my current thinking is to do it there.

Pike: I expect it to actually remove all the needs for `not(Entry)`

stasm: That would be a win. I'll try it, thanks!

stasm: I tried it out and it works well except that `Junk` doesn't now include trailing blank lines (if any). Which might actually make it a bit more like other `Entries`.

Pike: so if we'd join them afterwards, we'd have to pad our own newline? Or add top-level whitespace to the `Resource` production?

syntax/grammar.mjs
```diff
@@ -0,0 +1,496 @@
...
+            VariantKey.as("key")),
+        char("]"),
+        maybe(inline_space),
+        Pattern.as("value"))
+    .map(to_object)
+    .map(into(FTL.Variant))
+    .map(mutate({default: true})));
+
+let SelectorExpression = defer(() =>
```
Pike: I think this would read better if it was right after `SelectExpression`.

syntax/grammar.mjs
```diff
@@ -0,0 +1,496 @@
...
+    .map(into(FTL.SelectExpression)));
+
+let variant_list = defer(() =>
+    sequence(
+        repeat(Variant),
+        DefaultVariant,
+        repeat(Variant))
+    .map(flatten(1)));
+
+let VariantName = defer(() =>
```
Pike: `VariantName` and `VariantKey` threw me off here coming before `Variant` and `DefaultVariant`. Want to move them down?

syntax/grammar.mjs
```diff
@@ -0,0 +1,496 @@
...
+
+let MessageAttributeExpression = defer(() =>
+    sequence(
+        Identifier.as("id"),
+        char("."),
+        Identifier.as("name"))
+    .map(to_object)
+    .map(into(FTL.AttributeExpression)));
+
+let TermAttributeExpression = defer(() =>
```
Pike: Move this below `SelectorExpression`, and in particular out of inline expressions? As it's not one ;-)

Pike:
I think there are a couple of things that'd be good to put into the changelog, and probably also the final commit message. I'd call out

* explicit addition of `Junk`
* `Comment` being explicitly bound to `Message` and `Term` if adjacent
* obviously the shift of roles between adsl and mjs

I've made a pass through the ebnf in full now, the changes look good.

I'm still concerned about the plan to drop the ebnf. Not sure if that's my brain only, but I benefit significantly from the consice notion. I'll go through the exercise of reading grammar.mjs again tomorrow. Adding that comment here, because the role of the ebnf should also be part of the changelog entry, I guess.

syntax/grammar.mjs
stasm: Post-processing Junk changes how the following is parsed:

```
junk_line1\n
\n
junk_line2\n
\n
```

Before it was a single `Junk` entry which included all 4 `\n`. When we join junk lines in post-processing, those blank_lines don't make it there because they are dropped from the CST before `map(into(FTL.Resource))` is called. In a naive joining approach, the new single `Junk` would now only contain `junk_line1\njunk_line2`. Note the missing middle and trailing `\n`

One option is to let blank lines make it to the `FTL.Resource` construction and strip them there as the last post-processing step. This prevents the two junk lines above from being joined because there aren't adjacent now. The result is that we end up with _two_ `Junk` entries now, each for a contiguous block of junk lines.

I implemented this behavior in a commit which I'll push in a few.

Pike:
Now, onwards to the role of `grammar.mjs` as a reference parser.

I wonder if we should refactor `regular_char` and `text_char` alongside with `word` and `text_cont` to avoid the exceptions into an `either`. My guess is that many parser architectures will struggle to implement that as is, and we might be better off giving an example of how to implement it right?

Maybe with a full `ordenary_char` as `regular_char without any jazz`, and then just add our special chars explicitly where allowed?

Also, it might be easier to read if we added/removed character groups rather than single chars.

Side note, I wonder if we should have both `char()` and `string()`. I consider the difference to be an implementation detail, and the implementation of `string` is sadly one that I blame for many of the perf problems of the current python parser. The profile I had was showing deep stacks of single char checks. This being reference impl and all, we could just suggest to implement string checks as full string checks.

stasm:
> I'm still concerned about the plan to drop the ebnf. Not sure if that's my brain only, but I benefit significantly from the consice notion. I'll go through the exercise of reading grammar.mjs again tomorrow. Adding that comment here, because the role of the ebnf should also be part of the changelog entry, I guess.

I don't think we should drop it entirely, sorry if I didn't make it clear before. I think we should base our discussions on `grammar.mjs` because it allows ideas to be tested but continue to use the EBNF as quickly-look-something-up reference and as a easily digestible grammar cheatsheet.

(I didn't write the whole EBNF serialization just for the sake of this PR 😄. I want to keep using it going forward).

stasm:
Can we move all of your feedback from https://github.com/projectfluent/fluent/pull/103#issuecomment-389487454 re. `text_char`, `regular_char`, character classes and `string` vs. `char` into follow-ups? They are valuable improvements but I wouldn't like to block landing this.

lib/combinators.mjs
```diff
@@ -0,0 +1,139 @@
+import Parser from "./parser.mjs";
+import {Success, Failure} from "./result.mjs";
+import {join} from "./mappers.mjs";
+
+export function defer(fn) {
+    // Parsers might be defined as defer(() => parser) to avoid cyclic
+    // dependecies.
```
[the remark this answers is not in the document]
stasm: We still have a few parsers down at the bottom of `grammar.mjs` which don't use `defer`.

test/fixtures/select_expressions.ftl
```diff
@@ -0,0 +1,25 @@
...
+
+invalid-selector =
+    { -term[case] ->
+       *[key] value
+    }
+
+variant-list =
+    {
+       *[key] value
+    }
```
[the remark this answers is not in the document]
stasm: I'd prefer to stick to the current EBNF as much as possible, even if it means implementing something we already agreed should go away.

It's a bad practice to do this in a Message, yes. I'll add semantics checks later, after this lands. That's #99.

test/fixtures/escaped_characters.ftl
```diff
@@ -0,0 +1,5 @@
+backslash = Value with \\ (an escaped backslash)
+closing-brace = Value with \{ (a closing brace)
+unicode-escape = \u0041
+escaped-unicode = \\u0041
```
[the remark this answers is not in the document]
stasm: Ouch, thanks for catching that. I'll fix it by storing the escaped `\` in the AST.

stasm:
Thanks again, @Pike!

> I think that in 0.5, we allow for multline argument lists, with the infamous `__?` being an optional break-indent in https://github.com/projectfluent/fluent/blob/69ecabcc06cf6f5b3654c439a9defc8575744313/spec/fluent.ebnf#L77-L78.

Right, good catch. It wouldn't be beneficial to try to match the old EBNF because it's wrong. The `__?` doesn't allow for any inline space before the `break_indent`. I added a new `space_indent` production defined as `(inline_space break_indent?) | (inline_space? break_indent)` to fix this. I changed `CallExpression` and `Argument` to use it.

We should fix the same `__?` in `Placeable` but that's a bit harder because we don't want to consume the `break_indent` too quickly in case it's a `variant_list` (which must start with a `break_indent` too). I left only `inline_space` inside of `Placeable` for now. It's still an improvement over the old EBNF (which was broken) and we'll fix it in #102 by separating variant lists out of Placeables completely.

> The next issue is that `\u1234` and `\\u1234` result in the same AST. That's because the unescaping of `\` is assymetric between the unicode escape and `\\`, `\{`, etc. I suggest to not unescape at the syntax or ast level. Maybe we should have an AST node `Backtick`?

I agree with your earlier comment: let's store the special characters in their escaped form in the AST.

> I've also looked at the use of `element_at`, even before I hit the unescaping issue. In the non-escaping cases, would using `.abstract` and `.map(keep_abstract)` be more consistent than `element_at(2)`? This is used to skip whitespace etc in the AST a twice.

Because of the above, a simple `map(join)` does the trick :)

> Lastly, I hit a technical snag trying to check out different branches by a locally modified package-lock.json, in our case, I don't think we should commit it? Or actually pin versions so that local installs don't conflict?

`package-lock.json` isn't committed on master, so I'm not sure what kind of conflict you ran into. In any case, I think we _should_ commit it and yes, we should also pin versions of dependencies exactly (which we do) (unless I misunderstood?).

> And I forgot that I asked about the changelog entries, but maybe we do them in a follow-up.

I'll add it in this PR and then probably edit them again when publishing the 0.6 version. Thanks for the reminder!

syntax/grammar.mjs
```diff
@@ -0,0 +1,502 @@
...
+    .map(into(FTL.TextElement)));
+
+let Placeable = defer(() =>
+    sequence(
+        char("{"),
+        maybe(inline_space),
+        either(
+            // Order matters!
+            BlockExpression,
+            InlineExpression),
```
Pike: following up on an earlier comment, should this be `either(BE, IE).abstract` and s/element_at/keep_abstract/ ?

syntax/grammar.mjs
```diff
@@ -0,0 +1,502 @@
...
+    .map(join);
+
+let space_indent =
+    either(
+        sequence(
+            inline_space,
+            maybe(break_indent)),
+        sequence(
+            maybe(inline_space),
+            break_indent));
```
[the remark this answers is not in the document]
stasm: I'd like to keep using the `maybe()` combinator because it serializes as `?` in the EBNF. Your version of `space_indent` is how I originally implemented it :) I changed it to remove the redyndancy of nested `maybes`. If you're OK with it, I'll go with `maybe(space_indent)` and your implementation of `space_indent`.

syntax/grammar.mjs
Pike: sounds good. I had the feeling you'd like to keep the maybe() top-level ;-), the way it renders in the ebnf is a good supportive argument

stasm: Is this comment about the convention or is there some other reason to prefer `keep_abstract`? The reason why I used `element_at` here is that it returns a scalar value so I can then use `into` rather than `list_into` which made a little bit more sense to me semantically. There's only one thing that we care about inside of `Placeable`. I'm not overly attached to this reasoning, however, so I'm happy to change it if you feel it will improve the readability.

Issue #105, opened 2018-05-16
Explicit ordering of positional and named arguments in call expressions


Pike:
Right now, 

```
{ FOO("5", arg:4, 5) }
```

is a valid function call. Should the spec prescribe that named argument follow positional arguments?

Found this while reviewing #103.

stasm:
I don't think we can fix this in our grammar, but it should be possible to reject invalid orderings in `abstract.mjs`.

I tried editing the grammar to first expect `positional_arguments`, followed by optional `named_arguments`. The problem that I ran into is that as soon as we separate arguments into two types, the name of the first named argument parses as a valid positional `MessageReference`:

    FUN(a, b: 1)
            ^--- Positional arguments parse up to this point
                 and "b" has already been consumed when the parser
                 tries to parse named arguments.

I looked up how Python defines its argument lists in their [grammar](https://docs.python.org/3/reference/grammar.html). They also move the [validation of args into the AST construction](https://github.com/python/cpython/blob/1179f4b40f375af5c59cd4b6be9cc313fa0e1a37/Python/ast.c#L2765-L2778).

```
# Illegal combinations and orderings are blocked in ast.c:
# multiple (test comp_for) arguments are blocked; keyword unpackings
# that precede iterable unpackings are blocked; etc.
argument: ( test [comp_for] |
            test '=' test |
            '**' test |
            '*' test )
```

PR #112, opened 2018-05-21
Rename ExternalArgument to VariableReference

### LINKED ISSUE #100: Rename External Argument to Variable
I hear people refer to external arguments as to _variables_. I'd like to make it official. To quote Merriam-Webster:

  - _variable_ - a quantity that may assume any one of a set of values,
  - _variant_ - something that is different in some way from others of the same kind.

stasm:
Fix #100.


guide/functions.md
```diff
@@ -25,13 +25,10 @@ today-is = Today is { DATETIME($date) }
...
 ## Function parameters
 
-Functions may (but don't have to) accept positional and keyword arguments. Some
-keyword parameters are only available to developers and cannot be used inside
-of FTL translations. Instead they can be set on external arguments passed into
-translations by the developer.
-
-See the reference below for more information about the arguments accepted for
-each built-in function.
+Functions may accept positional and keyword arguments. Some keyword parameters
```
Pike: The docs are not up-to-date anymore here :-/

stasm:
I don't recall anything form yesterday about this. I know he said he liked _arguments_ from the perspective of the developer passing values into the translation and how _variables_ were something he'd expect to be able to assign to from inside of the localization. These are valid arguments (pun intended). I'm trying to look at this from the POV of the localizer who might not have had much experience with programming. however. `$num` is _variable_ (as in the adjective: _changing_) and because of its changing nature, `SelectExpressions` can select different _variants_. This wording has the benefit of avoiding the programmer's lingo of _passing arguments into_ other things.

guide/functions.md
stasm: Right, I missed this in #105. I'll fix it here since I'm already changing this line. Depending on the outcome of the discussion in #124, we might be able to fix this guide-wide there.

zbraniecki:
Does your change impact naming scheme in FluentDOM for example? We use id/args everywhere. Should we switch to `data-l10n-vars` or `data-l10n-refs`? :(

stasm:
Developer still pass variables as arguments to the translation, so I don't think a change is required in the bindings API. However, it would be nice to perhaps use `data-l10n-vars` for consistency. Fortunately, API changes should be much easier to make than syntax changes.

PR #113, opened 2018-05-21
Require named arguments to follow all positional arguments

### LINKED ISSUE #105: Explicit ordering of positional and named arguments in call expressions

stasm:
Fix #105.

### DIFF
--- a/syntax/abstract.mjs
+++ b/syntax/abstract.mjs
@@ -1,11 +1,25 @@
 import * as FTL from "./ast.mjs";
-import {always} from "../lib/combinators.mjs";
+import {always, never} from "../lib/combinators.mjs";
 
 export function list_into(Type) {
     switch (Type) {
         case FTL.CallExpression:
-            return ([callee, args = []]) =>
-                always(new Type(callee, args));
+            return ([callee, args]) => {
+                let positional = [];
+                let named = [];
+                for (let arg of args) {
+                    if (arg instanceof FTL.NamedArgument) {
+                        named.push(arg);
+                    } else if (named.length > 0) {
+                        return never(
+                            "Positional arguments must not follow " +
+                            "named arguments.");
+                    } else {
+                        positional.push(arg);
+                    }
+                }
+                return always(new Type(callee, positional, named));
+            };
         case FTL.Message:
             return ([comment, ...args]) =>
                 always(new Type(...args, comment));
--- a/syntax/abstract.mjs
+++ b/syntax/abstract.mjs
@@ -1,36 +1,38 @@
 import * as FTL from "./ast.mjs";
+import {always} from "../lib/combinators.mjs";
 
 export function list_into(Type) {
     switch (Type) {
         case FTL.CallExpression:
             return ([callee, args = []]) =>
-                new Type(callee, args);
+                always(new Type(callee, args));
         case FTL.Message:
             return ([comment, ...args]) =>
-                new Type(...args, comment);
+                always(new Type(...args, comment));
         case FTL.Pattern:
             return elements =>
-                new FTL.Pattern(
+                always(new FTL.Pattern(
                     elements
                         .reduce(join_adjacent(FTL.TextElement), [])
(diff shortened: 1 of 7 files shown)


syntax/abstract.mjs
```diff
@@ -1,36 +1,52 @@
 import * as FTL from "./ast.mjs";
+import {always, never} from "../lib/combinators.mjs";
```
stasm: This lays a foundation for adding more semantics checks to `abstract.mjs`. By making `into` return a new `Parser` instance itself, we can chain parsers (a.k.a. _flatmap_, `then` or `and_then`) and reject parse results during the AST construction.

stasm: Would these combinators read better as `accept` and `reject`?

syntax/ast.mjs
```diff
@@ -147,11 +147,12 @@ export class VariantExpression extends Expression {
 }
 
 export class CallExpression extends Expression {
-    constructor(callee, args = []) {
+    constructor(callee, args = [], kwargs = []) {
         super();
         this.type = "CallExpression";
         this.callee = callee;
         this.arguments = args;
+        this.keywords = kwargs;
```
stasm: The runtime has to separate positional and named arguments anyways, so we might just as well do it in the AST. I called the field `keywords` following how Python calls them. The arguments, however, are called _named arguments_, so perhaps there's some better wording here?

syntax/abstract.mjs
```diff
@@ -1,36 +1,52 @@
...
     switch (Type) {
         case FTL.CallExpression:
-            return ([callee, args = []]) =>
-                new Type(callee, args);
+            return ([callee, args]) => {
+                let posargs = [];
+                let kwargs = [];
+                for (let arg of args) {
+                    if (arg instanceof FTL.NamedArgument) {
+                        kwargs.push(arg);
```
stasm: In the future, we could add a semantic check here to forbid duplicate named arguments. I filed #114 to track this.

test/fixtures/call_expressions.json
```diff
@@ -35,7 +35,8 @@
                                         "name": "msg"
                                     }
                                 }
-                            ]
+                            ],
+                            "keywords": []
```
stasm: I re-read the whole file in full and verified that these changes make sense.

syntax/ast.mjs
Pike: I wouldn't introduce `keyword` in fluent. I'd make positional vs named explicit, as in `positional_arguments` and `keyword_arguments`. If that's too long, maybe `_args`, or just no suffix, and comment `// arguments`.

syntax/abstract.mjs
Pike: Those don't read significantly better to me. I wouldn't want two names for the same thing, though, and in grammar.mjs, the current terminology seems to be more fitting.

stasm:
> We might want to refactor abstract.mjs at some point down the road, but not yet. I think the point comes when there's more logic that validates, and it's getting hard to read the switch statement.

I agree. Let's first see how other changes planned for the 0.6 milestone affect this file. I'll be happy to refactor it before it gets unwieldy.

syntax/ast.mjs
stasm: What do you mean by `just no suffix, and comment // arguments.`? I also considered `options` for the named arguments, since they are mostly intended to be used as such.

Pike: No suffix meant to say
```
// arguments
positional = pos_args
named = named_args
```

stasm: Ah, OK! Yeah, I think I'll go for this one.

stasm: I'm going to land this PR with `positional` and `named`. I filed #124 to discuss renaming `named` to `options`.

Issue #97, opened 2018-04-25
Rules for using expressions in text and as selectors are too complex


stasm:
Terms can currently have values with variants as well as attributes. Variants can be accessed with `-term[variant name]` and attributes with `-term.attr`. Variants are supposed to be different facets of the same value (e.g. grammatical cases of the same noun) while attributes can provide additional grammatical information about the value, e.g. the gender, which can be used in selectors.

Messages can also define variant lists and attributes. However, accessing variants on messages is not allowed. Attributes can be access with `message.attr`.

To enforce this design we introduced rules for using the accessors in different contexts:

Expression|Allowed in text|Allowed as selector
----------|---------------|-------------------
`message`|Yes|No
`message.attr`|Yes|No
`message[variant]`<sup>1</sup>|—|—
`-term`|Yes|No
`-term.attr`|No|Yes
`-term[variant]`|Yes|No

> <sup>1</sup> Right now, this is not a valid expression.

The rules are complex. I'd like to make this part of the syntax design simpler.

stasm:
Relaxing these rules and making them linter checks would make it easier to write well-formed FTL.

zbraniecki:
When we designed it we did plan to enable `message[variant]` in the future, initially for a single-variantlist case, and if we see that being used `message[variant][subvariant]` or `message[variant, subvariant]` later.

stasm:
There still might be valid usecases for the `message[variant]` syntax and I'm not opposed to introducing it in the future. This issue to me is more about our stance towards grammar rules which govern which expressions may appear where. I'm leaning towards allowing all (currently supported) expressions everywhere where an expression may appear. Linters, rather than grammar, can protect against misuse and the runtime can use its fallback heuristics to ensure the translation will resolve to a string.

Pike:
I think I've come full circle on this once more (I'm getting dizzy ;-) )

I agree that removing options from accessors would make it easier to write well-formed FTL.

I do think that there's a level at which "simple" syntax helpers in editing modes and tools can help to write valid FTL.

Thinking about this in this context:

* messages don't have variants
* terms can have multi-depth variants (plurals, capitalization, case)
* attributes can have selectors (`PLATFORM()`-dependent gender), but no variants

I think we can have more powerful tools, and also human thinking, to help people to write working fluent, by having visual patterns and syntactic rules to assist.

[...]

PR #128, opened 2018-05-23
Validate Placeables and SelectExpressions in abstract.mjs

stasm:
…rather than in grammar.mjs.

### DIFF
--- a/lib/ebnf.mjs
+++ b/lib/ebnf.mjs
@@ -4,11 +4,12 @@ import visitor from "./visitor.mjs";
 import serialize from "./serializer.mjs";
 
 export default
-function ebnf(source) {
+function ebnf(source, min_name_length = 26) {
     let grammar_ast = babylon.parse(source, {sourceType: "module"});
     let rules = walk(grammar_ast, visitor);
     let state = {
         max_name_length: Math.max(
+            min_name_length,
             ...rules.map(rule => rule.name.length)),
     };
     return rules
--- a/syntax/abstract.mjs
+++ b/syntax/abstract.mjs
@@ -1,3 +1,11 @@
+/*
+ * AST Validation
+ *
+ * The parse result of the grammar.mjs parser is a well-formed AST which is
+ * validated according to the rules documented on the wiki:
+ * https://github.com/projectfluent/fluent/wiki/Syntax-Semantics
+ */
+
 import * as FTL from "./ast.mjs";
 import {always, never} from "../lib/combinators.mjs";
 
@@ -43,6 +51,15 @@ export function list_into(Type) {
                         .filter(remove_blank_lines)));
         case FTL.SelectExpression:
             return ([selector, variants]) => {
+                let invalid_selector_found =
+                    selector instanceof FTL.MessageReference
+                    || selector instanceof FTL.TermReference
+                    || selector instanceof FTL.VariantExpression
+                    || (selector instanceof FTL.AttributeExpression
+                        && selector.ref instanceof FTL.MessageReference);
+                if (invalid_selector_found) {
+                    return never("Invalid selector type: ${selector.type}.");
+                }
                 let invalid_variants_found = variants.some(
                     variant => variant.value instanceof FTL.VariantList);
                 if (invalid_variants_found) {
@@ -65,7 +82,22 @@ export function list_into(Type) {
 }
 
 export function into(Type) {
-    return (...args) => always(new Type(...args));
+    switch (Type) {
+        case FTL.Placeable:
+            return expression => {
+                let invalid_expression_found =
+                    expression instanceof FTL.AttributeExpression
+                    && expression.ref instanceof FTL.TermReference;
+                if (invalid_expression_found) {
+                    return never(
+                        "Invalid expression type: ${expression.type}.");
+                }
+                return always(new Type(expression));
+            };
+        default:
+            return (...args) =>
+                always(new Type(...args));
+    }
 }
 
 function join_adjacent(Type) {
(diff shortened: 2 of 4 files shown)


stasm:
This is something we talked about in #97. I was able to simplify the grammar and I added validity checks to `abstract.mjs`. This PR doesn't change anything wrt. the actual rules: what used to parse as junk, keeps doing so.

stasm:
My goal here is to make `grammar.mjs` describe a wider range of well-formed syntaxes, and specialize in `abstract.mjs`. This dual approach may allow some parsers to stick to a looser superset of well-formed grammar. For instance, the runtime parser would be able to skip the validity checks specified in `abstract.mjs` to optimize the parsing speed.

Additionally, moving validity checks into `abstract.mjs` gives us access to the children of the currently parsed production, which allows more complex rules tobe enforced.. This is harder to express in `grammar.mjs`. An example of this is #105.

zbraniecki:
Thanks for flagging me Axel!
It's a bit hard to forsee the impact of this change right now but, I'll give it a try:

> Does this make parsers easier to implement?

Yes.

> Does it improve error messages?

Yes.

> How does this play with editing modes? Would this make it easier or harder to write an autocompletion tool?

It should make it easier, because you can only do the pass-0 (`grammar.mjs`) while autocompleting, and only when the token is completed do pass-1 (`abstract.mjs`).

Now, saying that - this will not work for runtime script parsers, where we are CPU bound and want to do as few passes as possible. With compiled parsers like Rust, this is something that we will need to test out.
I'd like to try to work with my Kekoa to get a `nom` based Fluent parser and I hope that the double-pass is not going to affect performance in any noticable way.

HTH

stasm:
I'll wait for #129 to land and I'll then rebase this PR on top of it, and re-submit for review.

stasm:
@Pike Do you want to be on the review path for this? Previously, you said it looked OK from the technical POV. With @zbraniecki's comment, is this good to land?

Pike:
I'm still torn on this.

> It should make it easier, because you can only do the pass-0 (grammar.mjs) while autocompleting, and only when the token is completed do pass-1 (abstract.mjs).

I've ran around a bit to see if there's a pass-1 in editors. I guess that fluent.vim doesn't even try to do pass-0 for realz, right? I've spent some time trying to do pass-0 for Atom (updating tree-sitter-fluent), but also hit a point where I'm seeing weird parse errors that I need more time to investigate. I'd expect that we had more power in an mls-based editor, but that's speculation. I haven't even read the documentation to that extent.

> Now, saying that - this will not work for runtime script parsers, where we are CPU bound and want to do as few passes as possible.

That would mean that we need to ensure some of the pass-1 constraints in the resolver then, right? We can't generally assume that every fluent file went through an error-stripping build process?

lib/ebnf.mjs
```diff
@@ -4,11 +4,12 @@ import visitor from "./visitor.mjs";
 import serialize from "./serializer.mjs";
 
 export default
-function ebnf(source) {
+function ebnf(source, min_name_length = 26) {
```
Pike: Is this just to keep the diff small? Do we intend to keep this?

syntax/abstract.mjs
```diff
@@ -43,6 +43,15 @@ export function list_into(Type) {
                         .filter(remove_blank_lines)));
         case FTL.SelectExpression:
             return ([selector, variants]) => {
+                let invalid_selector_found =
+                    selector instanceof FTL.MessageReference
+                    || selector instanceof FTL.TermReference
+                    || selector instanceof FTL.VariantExpression
+                    || (selector instanceof FTL.AttributeExpression
+                        && selector.ref instanceof FTL.MessageReference);
```
Pike: I found this hard to digest. I wonder if a positive criteria is revealing the intent here better? I.e, instead of excluding `MessageReference`, include `StringLiteral`? That would also allow to return early.

It's yet one more condition, though. Also a reason why I'm still torn on the outside-impact on this change.

lib/ebnf.mjs
stasm: Yes, it's to minimize the diff. I don't mind leaving it here and giving the EBNF some breathing air. I also don't mind removing this change and using `git diff --ignore-space-change` (a.k.a. `git diff -b`) to view changes between syntax versions. Even GitHub has this option now, so maybe that's the way to go?

stasm:
> That would mean that we need to ensure some of the pass-1 constraints in the resolver then, right? We can't generally assume that every fluent file went through an error-stripping build process?

That's a really good question, thanks for asking.

The runtime parser is a stop-gap solution which allows us to ship Fluent in Firefox today, before we can use a Rust parser conveniently. The Mozilla products are the main driver of Fluent right now, and being able to leverage the existing localization infrastructure at Mozilla to provide a safety net for the runtime parser is an important part of the design of the runtime parser. In other words, it's OK for the runtime parser to only do the pass-0 parsing (or even something less strict) because it's mostly used in projects where the pass-1 validation happens in the tools (Pontoon) and the build system (compare-locales) anyways.

By the time Fluent is popular outside of Mozilla, we'll be able to use a full-featured parser written in Rust, likely compiled to WASM. It will perform both the pass-0 (`grammar.mjs`) and the pass-1 (`abstract.mjs`) validation.

syntax/abstract.mjs
stasm: I'd prefer to keep it as is for consistency with other rules. The rules highlight the invalid cases which are generally less frequent.

Pike:
It'd be nice to see how much of abstract.mjs is actually caught in the runtime resolver.

zbraniecki:
I fully agree with what @stasm wrote above.

stasm:
Thanks. When I talked to @Pike on Friday, it turned out that I misunderstood his comment a bit. I read it as talking about the runtime parser whereas @Pike meant the runtime resolver. Looking at the resolver's code it's somewhat similar to `abstract.mjs` in that it works on a well-formed AST. I agree with @Pike that there's an opportunity to try to enforce some validation rules there. It shouldn't add much code to the resolver. I think we should do it outside of the Syntax 0.6 implementation scope.

PR #129, opened 2018-05-24
Introduce the VariantList value type

### LINKED ISSUE #102: New value type: Variant List

stasm:
This PR removes selector-less `SelectExpressions` in favor of `VariantLists` which share the same syntax. Values of `Messages` and `Terms` may now be either `Patterns` or `VariantLists`. Values of `Variants` may only be `VariantLists` if they're defined in another `VariantList`.

Fixes #102.

### DIFF
--- a/syntax/abstract.mjs
+++ b/syntax/abstract.mjs
@@ -36,9 +36,23 @@ export function list_into(Type) {
                     entries
                         .reduce(join_adjacent(FTL.Junk), [])
                         .filter(remove_blank_lines)));
+        case FTL.SelectExpression:
+            return ([selector, variants]) => {
+                let invalid_variants_found = variants.some(
+                    variant => variant.value instanceof FTL.VariantList);
+                if (invalid_variants_found) {
+                    return never(
+                        "VariantLists are only allowed inside of " +
+                        "other VariantLists.");
+                }
+                return always(new Type(selector, variants));
+            };
         case FTL.Term:
             return ([comment, ...args]) =>
                 always(new Type(...args, comment));
+        case FTL.VariantList:
+            return ([variants]) =>
+                always(new Type(variants));
         default:
             return elements =>
                 always(new Type(...elements));
--- a/syntax/grammar.mjs
+++ b/syntax/grammar.mjs
@@ -55,7 +55,7 @@ let Term = defer(() =>
         maybe(inline_space),
         char("="),
         maybe(inline_space),
-        Pattern.abstract,
+        Value.abstract,
         repeat(Attribute).abstract,
         line_end)
     .map(keep_abstract)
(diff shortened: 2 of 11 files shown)


syntax/abstract.mjs
```diff
@@ -36,9 +36,18 @@ export function list_into(Type) {
                     entries
                         .reduce(join_adjacent(FTL.Junk), [])
                         .filter(remove_blank_lines)));
+        case FTL.SelectExpression:
+            return ([selector, variants]) =>
+                variants.some(variant =>
+                    variant.value instanceof FTL.VariantList)
+                    ? never("")
```
Pike: Please add an actual error message here? Maybe `Select expressions can only have patterns or select expressions as values` ?

stasm: Sorry, I missed this after I resolved a conflict with #128.

stasm:
I edited the [Syntax Semantics](https://github.com/projectfluent/fluent/wiki/Syntax-Semantics) wiki page to add rules for validating `VariantLists`. I'll resubmit this PR on Monday.

syntax/grammar.mjs
```diff
@@ -111,19 +111,33 @@ let Attribute = defer(() =>
         maybe(inline_space),
         char("="),
         maybe(inline_space),
-        Pattern.abstract)
+        Value.abstract)
```
Pike: I think we should undo this part, attributes in general don't have an `XAttributeVariantExpression`, independent of Message or Term.

stasm: Good point, thanks. I think #128 will make this easier to add in the future, because the grammar won't need to enumerate each of the possible `X` in `X{Attribute,Variant}Expression`.

stasm: In fact, I could also keep the `Pattern` in `Message`, too. Do you have a preference wrt. checking the values of Messages and Attributes in `grammar` vs. in `abstract`?

Pike: I agree. I think `grammar` would be a good place. That makes the variants for terms stick out more.

test/fixtures/variant_lists.json
```diff
@@ -0,0 +1,225 @@
...
+            },
+            "attributes": [],
+            "comment": null
+        },
+        {
+            "type": "Term",
+            "annotations": [],
+            "id": {
+                "type": "Identifier",
+                "name": "-variant-list-in-term-attr"
```
stasm: This `Term` parsed fine up to the broken `Attribute`. I filed #132 to discuss the desired behavior of the parser in such cases.

test/fixtures/variant_lists.json
```diff
@@ -0,0 +1,225 @@
...
+            "type": "Junk",
+            "annotations": [],
+            "content": "variant-list-in-message =\n    {\n       *[key] Value\n    }\n"
+        },
+        {
+            "type": "Message",
+            "annotations": [],
+            "id": {
+                "type": "Identifier",
+                "name": "variant-list-in-message-attr"
```
stasm: Idem for the `Message` here.

stasm:
I made the changes you requested in a new commit, which I'll squash before landing. `Message` and `Attribtue` values can only be `Patterns` which is enforced in the grammar. This made changes to `abstract` simpler; it only checks `Variant` values inside of `SelectExpressions`.

PR #137, opened 2018-05-30
Initial syntax hightlighting for vim/neovim

zbraniecki:
While waiting for Firefox to recompile I put together some very basic syntax highlighting for vim. It does decent job when looking at .ftl files in `mozilla-central` at least and maybe be a good foundation.

Not sure if you want to handle those in tree or in gist all around or some separate repo, but I think the easiest location would be in this repo so placing it as a PR.

### DIFF
--- a/editors/vim/syntax/fluent.vim
+++ b/editors/vim/syntax/fluent.vim
@@ -1,0 +1,26 @@
+if exists("b:current_syntax")
+    finish
+endif
+
+syntax match fluentComment "\v#.*$"
+syntax match fluentIdentifier "^\v[a-zA-Z-]+" nextgroup=fluentDelimiter
+syntax match fluentDelimiter "\s*=\s*" contained skipnl nextgroup=fluentPattern
+syntax region fluentPattern contained start="" end="\v^(( )@!|( +[\.\[\*\}])@=)" contains=fluentPlaceableExpression
+syntax match fluentAttribute "\v\.[a-zA-Z-]+" nextgroup=fluentDelimiter
+syntax region fluentPlaceableExpression contained matchgroup=fluentPlaceableBraces start=+{+ end=+}+ contains=@fluentExpression
+
+syntax cluster fluentExpression contains=fluentBuiltin,fluentVariantEntry,fluentIdentifierExpression
+
+syntax match fluentBuiltin contained "\v[A-Z]+(\()@="
+syntax match fluentVariantEntry contained "\v\*?\[[a-z]+\]" nextgroup=fluentPattern
+syntax match fluentIdentifierExpression contained "\v[a-zA-Z-]+"
+
+highlight link fluentComment Comment
+highlight link fluentIdentifier Identifier
+highlight link fluentPattern String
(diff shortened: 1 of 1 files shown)


editors/vim/syntax/fluent.vim
```diff
@@ -0,0 +1,26 @@
+if exists("b:current_syntax")
+    finish
+endif
+
+syntax match fluentComment "\v#.*$"
+syntax match fluentIdentifier "^\v[a-zA-Z-]+" nextgroup=fluentDelimiter
```
stasm: Can you make this a bit more like the actual `Identifier` grammar? In particular, allowing digits in non-first position is important for versioned strings.

editors/vim/syntax/fluent.vim
```diff
@@ -0,0 +1,26 @@
...
+syntax match fluentDelimiter "\s*=\s*" contained skipnl nextgroup=fluentPattern
+syntax region fluentPattern contained start="" end="\v^(( )@!|( +[\.\[\*\}])@=)" contains=fluentPlaceableExpression
+syntax match fluentAttribute "\v\.[a-zA-Z-]+" nextgroup=fluentDelimiter
+syntax region fluentPlaceableExpression contained matchgroup=fluentPlaceableBraces start=+{+ end=+}+ contains=@fluentExpression
```
stasm: Can this be simply called `fluentPlaceable`? Also, what's `fluentPlaceableBrace` and where is it used/defined?

editors/vim/syntax/fluent.vim
```diff
@@ -0,0 +1,26 @@
...
+
+syntax cluster fluentExpression contains=fluentBuiltin,fluentVariantEntry,fluentIdentifierExpression
+
+syntax match fluentBuiltin contained "\v[A-Z]+(\()@="
```
stasm: Please call this `fluentFunction`.

editors/vim/syntax/fluent.vim
```diff
@@ -0,0 +1,26 @@
...
+syntax match fluentVariantEntry contained "\v\*?\[[a-z]+\]" nextgroup=fluentPattern
```
stasm: Please call this `fluentVariantKey` and allow digits, too.

editors/vim/syntax/fluent.vim
```diff
@@ -0,0 +1,26 @@
...
+syntax match fluentIdentifierExpression contained "\v[a-zA-Z-]+"
```
stasm: Is this a `MessageReference`? Can you rename it if so, and also add `TermReference` and `VariableReference`? I think these are the most common ones.

Pike:
FWIW, I agree with "own repo". We should also look at how our atom support evolves in parallel (still owe Max feedback).

For fluent, I also find indent scripts interesting, http://psy.swansea.ac.uk/staff/carter/vim/vim_indent.htm has some hints on how to write those.

zbraniecki:
Thanks! Landed as https://github.com/projectfluent/fluent.vim

PR #146, opened 2018-06-06
Support just running an individual test

Pike:
I found this nice as I was writing tests for a dedicated feature and just wanted to trace one test instead of all.

### DIFF
--- a/test/parser.mjs
+++ b/test/parser.mjs
@@ -7,16 +7,24 @@ const fixtures_dir = process.argv[2];
 
 if (!fixtures_dir) {
     console.error(
-        "Usage: node --experimental-modules parser.mjs FIXTURES_DIR");
+        "Usage: node --experimental-modules parser.mjs FIXTURES_DIR_OR_FTL");
     process.exit(1);
 }
 
 main(fixtures_dir);
 
 async function main(fixtures_dir) {
-    const files = await readdir(fixtures_dir);
-    const ftls = files.filter(
-        filename => filename.endsWith(".ftl"));
+    let files, ftls;
+    if (fixtures_dir.endsWith(".ftl")) {
+        // Actually, this is a filepath, split the path and the dir.
+        ftls = [path.basename(fixtures_dir)];
+        fixtures_dir = path.dirname(fixtures_dir);
+    }
+    else {
+        files = await readdir(fixtures_dir);
+        ftls = files.filter(
+            filename => filename.endsWith(".ftl"));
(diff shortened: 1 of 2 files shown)


test/parser.mjs
```diff
@@ -7,16 +7,24 @@ const fixtures_dir = process.argv[2];
...
 main(fixtures_dir);
 
 async function main(fixtures_dir) {
-    const files = await readdir(fixtures_dir);
-    const ftls = files.filter(
-        filename => filename.endsWith(".ftl"));
+    let files, ftls;
+    if (fixtures_dir.endsWith(".ftl")) {
+        // Actually, this is a filepath, split the path and the dir.
+        ftls = [path.basename(fixtures_dir)];
```
stasm: This looks like a good place to use a `var`! You can declare `ftls` as `var` both here and inside of the `else` block.

test/parser.mjs
```diff
@@ -7,16 +7,24 @@ const fixtures_dir = process.argv[2];
...
+        fixtures_dir = path.dirname(fixtures_dir);
+    }
+    else {
+        files = await readdir(fixtures_dir);
```
stasm: `files` can be block-scoped here with `let`.

test/parser.mjs
```diff
@@ -7,16 +7,24 @@ const fixtures_dir = process.argv[2];
 
 if (!fixtures_dir) {
     console.error(
-        "Usage: node --experimental-modules parser.mjs FIXTURES_DIR");
+        "Usage: node --experimental-modules parser.mjs FIXTURES_DIR_OR_FTL");
```
stasm: Maybe just call this a `FIXTURE`?

test/parser.mjs
```diff
@@ -7,16 +7,24 @@ const fixtures_dir = process.argv[2];
...
-    const files = await readdir(fixtures_dir);
-    const ftls = files.filter(
-        filename => filename.endsWith(".ftl"));
+    let files, ftls;
+    if (fixtures_dir.endsWith(".ftl")) {
+        // Actually, this is a filepath, split the path and the dir.
+        ftls = [path.basename(fixtures_dir)];
+        fixtures_dir = path.dirname(fixtures_dir);
+    }
+    else {
```
stasm: Nit: put this on the same line as `}`.

stasm:
@Pike Feel free to squash and land this yourself, or let me know if you'd like me to do it.

PR #147, opened 2018-06-06
unify char and string, just use char as it's the smaller diff, fixes #108

### LINKED ISSUE #108: Optimize string parsing
In https://github.com/projectfluent/fluent/pull/103#issuecomment-389487454 @Pike writes:

> Side note, I wonder if we should have both `char()` and `string()`. I consider the difference to be an implementation detail, and the implementation of `string` is sadly one that I blame for many of the perf problems of the current python parser. The profile I had was showing deep stacks of single char checks. This being reference impl and all, we could just suggest to implement string checks as full string checks.

Pike:
This patch adresses my comments that ended up being issue #108.

I did run this patch through bench, didn't matter much.

I started digging into this as I hacked on the backports branch, and it took me hours to figure out why `char("//")` doesn't work. Cause that also just silently failed :-)

### DIFF
--- a/lib/stream.mjs
+++ b/lib/stream.mjs
@@ -8,7 +8,7 @@ export default class Stream {
     }
 
     // Get the element at the cursor.
-    head() {
+    head(len=1) {
         if (this.length < 0) {
             return undefined;
         }
@@ -17,7 +17,10 @@ export default class Stream {
             return Symbol.for("eof");
         }
 
-        return this.iterable[this.cursor];
+        if (len === 1) {
+            return this.iterable[this.cursor];
+        }
+        return this.iterable.slice(this.cursor, this.cursor + len);
     }
 
     // Execute a regex on the iterable.
--- a/lib/stream.mjs
+++ b/lib/stream.mjs
@@ -17,9 +17,6 @@ export default class Stream {
             return Symbol.for("eof");
         }
 
-        if (len === 1) {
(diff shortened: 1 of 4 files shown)


lib/stream.mjs
```diff
@@ -17,7 +17,10 @@ export default class Stream {
             return Symbol.for("eof");
         }
 
-        return this.iterable[this.cursor];
+        if (len === 1) {
+            return this.iterable[this.cursor];
```
stasm: Is this a perf optimization? Can we leave it out?

lib/stream.mjs
```diff
@@ -8,7 +8,7 @@ export default class Stream {
     }
 
     // Get the element at the cursor.
-    head() {
+    head(len=1) {
```
stasm: Nit: JS style is to use spaces around the `=` here.

syntax/grammar.mjs
```diff
@@ -282,7 +282,7 @@ let SelectExpression = defer(() =>
     sequence(
         InlineExpression.abstract,
         maybe(inline_space),
-        string("->"),
+        char("->"),
```
stasm: Looking at the usage here and in other places, shouldn't we keep `string` (or `str`) rather than `char`?

lib/stream.mjs
Pike: Adding a config to eslint to tell me next time.

PR #148, opened 2018-06-07
Back port pre 0.5

Pike:
This is the start of a tooling-legacy-whatchamacallit branch.

I've chosen `pre-0.5` as code name instead of `0.4`, because I figured that's more adequate? Like, 0.3 had these features, too.

@stasm, not really ready for review yet, but what do you think about this?

### DIFF
--- a/spec/fluent.ebnf
+++ b/spec/fluent.ebnf
@@ -3,12 +3,15 @@
 Resource            ::= (blank_line | Entry | junk_line)*
 Entry               ::= Message
                       | Term
-                      | (ResourceComment | GroupComment | Comment)
-Message             ::= Comment? Identifier inline_space? "=" inline_space? ((Pattern Attribute*) | (Attribute+)) line_end
+                      | (ResourceComment | GroupComment | Comment | LegacyComment)
+Message             ::= (Comment | LegacyComment)? Identifier inline_space? "=" inline_space? ((Pattern Attribute*) | (Attribute+)) line_end
 Term                ::= Comment? TermIdentifier inline_space? "=" inline_space? Value Attribute* line_end
 Comment             ::= ("#" comment_line)+
 GroupComment        ::= ("##" comment_line)+
 ResourceComment     ::= ("###" comment_line)+
+
+/* "//"=style comments were removed in 0.5 */
+LegacyComment       ::= ("//" comment_line)+
 
 /* Adjacent junk_lines should be joined into FTL.Junk during the AST
    construction. */
--- a/spec/fluent.ebnf
+++ b/spec/fluent.ebnf
@@ -1,114 +1,119 @@
 
 /* An FTL file defines a Resource. */
-Resource            ::= (blank_line | Entry | junk_line)*
-Entry               ::= Message
-                      | Term
-                      | LegacySection
-                      | (ResourceComment | GroupComment | Comment | LegacyComment)
-Message             ::= (Comment | LegacyComment)? Identifier inline_space? "=" inline_space? ((Pattern Attribute*) | (Attribute+)) line_end
-Term                ::= Comment? TermIdentifier inline_space? "=" inline_space? Value Attribute* line_end
-Comment             ::= ("#" comment_line)+
-GroupComment        ::= ("##" comment_line)+
-ResourceComment     ::= ("###" comment_line)+
+Resource               ::= (blank_line | Entry | junk_line)*
+Entry                  ::= Message
+                         | Term
+                         | LegacyAttributeMessage
+                         | LegacySection
+                         | (ResourceComment | GroupComment | Comment | LegacyComment)
+Message                ::= (Comment | LegacyComment)? Identifier inline_space? "=" inline_space? ((Pattern Attribute*) | (Attribute+)) line_end
+Term                   ::= Comment? TermIdentifier inline_space? "=" inline_space? Value Attribute* line_end
+Comment                ::= ("#" comment_line)+
+GroupComment           ::= ("##" comment_line)+
+ResourceComment        ::= ("###" comment_line)+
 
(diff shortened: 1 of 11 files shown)


syntax/ast.mjs
```diff
@@ -56,6 +56,14 @@ export class Term extends Entry {
     }
 }
 
+export class Section extends SyntaxNode {
+    constructor(words) {
+        super();
+        this.type = "Section";
+        this.words = words;
```
stasm: This should be called `name`.

spec/fluent.ebnf
```diff
@@ -3,13 +3,17 @@
...
                       | (ResourceComment | GroupComment | Comment)
 Message             ::= Comment? Identifier inline_space? "=" inline_space? ((Pattern Attribute*) | (Attribute+)) line_end
 Term                ::= Comment? TermIdentifier inline_space? "=" inline_space? Value Attribute* line_end
-Comment             ::= ("#" comment_line)+
+Comment             ::= (("#" | "//") comment_line)+
 GroupComment        ::= ("##" comment_line)+
 ResourceComment     ::= ("###" comment_line)+
 
+/* Sections got removed in 0.5 */
+Section             ::= "[[" inline_space? word (inline_space? word)* inline_space? "]]"
```
stasm: I suggest adding a `section_name` production and using it here. It will simplify the `map()` operations in `grammar.mjs` for `Section`.

syntax/ast.mjs
Pike: Fixed.

spec/fluent.ebnf
Pike: Is trying for 2 hours and giving up a good-enough reason to not do this? Once this is further along, I might also take a follow-up, if that's easy for you to do. I just can't get the data out after copying and pasting working code :-(

Pike:
> Do you also plan to add the support for the optional = in value-less messages?

Yes, I just didn't want to do the full stack and then rename everything.

And yes, this is the fork branch, that I then intend to merge into a stable branch. `archaeology` sounds OK, aside from being hard to remember when trying to spell it :bee:

spec/fluent.ebnf
stasm: Sure, I can land a follow up afterwards.

syntax/grammar.mjs
```diff
@@ -64,7 +66,10 @@ let Term = defer(() =>
 let Comment = defer(() =>
     repeat1(
         sequence(
-            string("#"),
+            either(
+                string("#"),
+                string("//"), // pre 0.5
```
stasm: It would be nice to have this comment in the generated EBNF, too. Maybe even add a new production dedicated to the legacy `//` comment? Although this would likely complicate the definition of `Message` and `Term`, so maybe it's not worth it.

stasm: Argh, re-reading, I realize it might not be clear what `comment` refers to above. I'll try to rephrase:

> It would be nice to have this comment (i.e. `// pre 0.5`) in the generated EBNF, too. Maybe even add a new production dedicated to the legacy pre-0.5 Comment?

Pike: Wasn't too hard, and I did a complete rebase-history thing, as I think the individual commits are good to take in the end. And reviewing the follow-up commit would have been hard, I guess.

spec/fluent.ebnf
```diff
@@ -3,13 +3,24 @@
 Resource            ::= (blank_line | Entry | junk_line)*
 Entry               ::= Message
                       | Term
-                      | (ResourceComment | GroupComment | Comment)
-Message             ::= Comment? Identifier inline_space? "=" inline_space? ((Pattern Attribute*) | (Attribute+)) line_end
+                      | Section
+                      | ValueLessMessage
```
stasm: I like `LegacyComment` and I wonder if in a similar vein we could have `LegacySection` and `LegacyMessage` (or `LegacyValueless`) here?

Just a suggestion, the current names LGTM as well, with a nit about `ValueLess`: should it be spelled `Valueless` with a lowercase `L`?

spec/fluent.ebnf
```diff
@@ -3,13 +3,24 @@
...
+                      | (ResourceComment | GroupComment | Comment | LegacyComment)
+Message             ::= (Comment | LegacyComment)? Identifier inline_space? "=" inline_space? ((Pattern Attribute*) | (Attribute+)) line_end
 Term                ::= Comment? TermIdentifier inline_space? "=" inline_space? Value Attribute* line_end
```
stasm: You're not adding `| LegacyComment` here because Terms weren't around pre-0.5? Nice.

spec/fluent.ebnf
```diff
@@ -3,13 +3,24 @@
...
 Comment             ::= ("#" comment_line)+
 GroupComment        ::= ("##" comment_line)+
 ResourceComment     ::= ("###" comment_line)+
 
+/* "//"=style comments got removed in 0.5 */
```
stasm: A stylistic nit of a non-native speaker: `were removed` reads better to me, even though I'm not really sure which one is better English.

spec/fluent.ebnf
```diff
@@ -3,13 +3,24 @@
...
+LegacyComment       ::= ("//" comment_line)+
+
+/* Sections got removed in 0.5 */
+Section             ::= "[[" inline_space? word (inline_space? word)* inline_space? "]]"
+
+/* Messages without value had no "=" prior to 0.5 */
+ValueLessMessage    ::= Comment? Identifier inline_space? (Attribute+) line_end
```
stasm: I think this needs `(Comment | LegacyComment)?`

Pike:
Yep, closing.

PR #151, opened 2018-07-05
Trim break_indent of inline space and only keep the newlines

### LINKED ISSUE #150: multi-line patterns should not include indention in value
This seems to be a regression in the 0.6 work.

For a

```
message = something
  indented
```

the leading whitespace is now part of the value, which it wasn't in 0.5 in practice.

I think that in particular in the scope of 0.6, we shouldn't make such changes.

stasm:
Fixes #150.

### DIFF
--- a/syntax/grammar.mjs
+++ b/syntax/grammar.mjs
@@ -479,7 +479,8 @@ let break_indent =
         repeat(blank_line),
         inline_space)
     .map(flatten(1))
-    .map(join);
+    // Trim the indents. Keep only the newlines.
+    .map(indents => indents.fill("").join("\n"));
 
 let space_indent =
     sequence(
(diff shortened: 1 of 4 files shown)


stasm:
@Pike This PR brings back the desired behavior from 0.5. It's also what `fluent-syntax` implements right now.

stasm:
> I like this, but trim is a tad too eager (it'd strip nbsp etc).

Is it, though? `break_indent` is defined strictly in terms of `line_end` and `inline_space` which means only the following characters are valid inside of it: ` ` (space), `\t`, `\n` and `\r`. I used `trim` to reduce each element of the array to an empty string so that I can join them with `\n` in the next step. I could have just as well written `.map(indents => indents.map(() => ""))` to achieve the same end effect. I chose `trim` to make it more clear what this step does.

stasm:
Or even:

```js
.map(indents => indents.fill("").join("\n"))
```

Would that be more clear?

stasm:
I went for the `fill()` approach.

PR #153, opened 2018-07-18
Don't include the trailing line break in Messages, Terms and Comments

stasm:
Revert earlier changes which added the trailing newline to the
definition of the Message, Term and Comment, as well as to the content
of Comments. The trailing line break should not be part of the
production, span nor the content.

Make Junk subclass SyntaxNode directly, rather than Entry. Entry is now
an abstract class for useful AST nodes carrying content: Messages, Terms
and Comments.

Remove the annotations field from Entry togather with the add_annotation
method and move it directly onto Junk.

### DIFF
--- a/syntax/abstract.mjs
+++ b/syntax/abstract.mjs
@@ -82,6 +82,25 @@ export function list_into(Type) {
 
 export function into(Type) {
     switch (Type) {
+        case FTL.Comment:
+        case FTL.GroupComment:
+        case FTL.ResourceComment:
+            return content => {
+                if (content.endsWith("\n")) {
+                    if (content.endsWith("\r\n")) {
+                        var offset = -2;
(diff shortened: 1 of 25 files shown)


syntax/abstract.mjs
```diff
@@ -82,6 +82,14 @@ export function list_into(Type) {
 
 export function into(Type) {
     switch (Type) {
+        case FTL.Comment:
+        case FTL.GroupComment:
+        case FTL.ResourceComment:
+            return content =>
+                // Only trim if the comment ends with an EOL, not EOF.
+                content.endsWith("\n")
```
Pike: We should do all three line-ending-variants here?

stasm: Ah, good point, thanks!

PR #157, opened 2018-07-26
Restrict variant keys to Identifiers and NumberLiterals

### LINKED ISSUE #127: Restrict variant keys to Numbers and Identifier
This is related to #90, in which I'd like to relax the grammar of variant keys. I'd like to start by _restricting it_ to just Numbers and Identifiers. We can relax this later and the benefit of restricting it now is to clarify the exact rules of parsing the surrounding whitespace, as in `[____key____]`.

stasm:
Fix #127.


stasm:
Hey @Pike, this implements my proposal from #127. I'd like to play it safe and restrict the grammar where it makes sense.

PR #161, opened 2018-08-03
Remove support for CR

### LINKED ISSUE #154: Remove support for CR as EOL
LF and CRLF make sense. CR died a decade ago together with Mac OS 9.

stasm:
Fix #154.

The second commit normalizes the parsing of all supported line endings to `LF`. We already use `LF` to represent blank lines in multiline patterns. Normalizing line endings in the parsed results removes surprises and ensures consistency across platforms. Serializers can implement an option to serialize using `CRLF` if so desired.

In essence, this is similar to git's [`core.autocrlf = input`](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_code_core_autocrlf_code) option.

### DIFF
--- a/spec/CHANGELOG.md
+++ b/spec/CHANGELOG.md
@@ -3,8 +3,7 @@
 
 ## Unreleased
 
-  - …
-
+  - Remove support for CR as a line ending. (#154)
 
 ## 0.6.0 (July 24, 2018)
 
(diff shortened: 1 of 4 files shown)


Pike:
I'd like to take your offer to handle the CR-removal and the newline-normalization in two different issues.

The CR one is pretty simple, the line ending died long before we introduced fluent. There's __maybe__ a tad of "I can use my old java properties files, 'cause they work", but really, no.

The newline normalization is something that deserves a deeper look, and maybe also a bit more external input. That seems like something in scope of an RFC.

stasm:
Sounds good!

stasm:
I removed the second commit from this PR and I'll file an RFC on Monday.

stasm:
I filed #163 to discuss the normalization of EOLs.

PR #166, opened 2018-08-15
Add test for white-space-only lines in values

Pike:
Add a test to codify how we're handling pattern lines with
just indention and completely empty lines.

I.e., both generate an empty line in the message value.

I don't think we have such a test yet, and with all the refactors, I think it'd be good to have. In particular when writing actual implementations based on the reference fixtures.

### DIFF
--- a/test/fixtures/whitespace_in_value.ftl
+++ b/test/fixtures/whitespace_in_value.ftl
@@ -2,6 +2,9 @@
 key =
   first line
 
+
   
+  
+
 
   last line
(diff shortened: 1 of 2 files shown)


test/fixtures/whitespace_in_value.ftl
```diff
@@ -0,0 +1,10 @@
+# Caution, lines 4 and 6 contain white-space-only lines
```
stasm: Looking at the diff, aren't they lines 6 and 7?

stasm: image

stasm:
Thanks for writing the test. Feel free to merge this in.

PR #167, opened 2018-08-17
Forbid tabs as inline blanks

### LINKED ISSUE #165: Use only spaces for syntax whitespace, drop tab for that
Modern languages strongly enforce either spaces or tabs for syntactic white-space.

We should do the same for Fluent, I think.

Given our white-space isn't just for indention, I think we should settle for space characters, and throw the tab into the pool of regular text content.

stasm:
Fix #165.

### DIFF
--- a/test/fixtures/tab.ftl
+++ b/test/fixtures/tab.ftl
@@ -1,0 +1,14 @@
+# Error (tab after =)
+key01 =	Value 01
+
+# Error (tab before =)
+key02	= Value 02
+
+# Error (tab is not a valid indent)
+key03 =
+	This line isn't properly indented.
+
+# Partial Error (tab is not a valid indent)
+key04 =
+    This line is indented by 4 spaces,
+	whereas this line by 1 tab.
--- a/spec/fluent.ebnf
+++ b/spec/fluent.ebnf
@@ -95,7 +95,7 @@ quoted_text_char    ::= (text_char - quote)
 digit               ::= [0-9]
 
 /* Whitespace */
-blank_inline        ::= ("\u0020" | "\u0009")+
+blank_inline        ::= "\u0020"+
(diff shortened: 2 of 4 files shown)


test/fixtures/tab.ftl
```diff
@@ -0,0 +1,14 @@
...
+key02	= Value 02
+
+# Error (tab is not a valid indent)
+key03 =
+	This line isn't properly indented.
+
+# Partial Error (tab is not a valid indent)
+key04 =
+    This line is indented by 4 spaces,
+	whereas this line by 1 tab.
```
stasm: This exhibits the same behavior as when a broken attribute is encountered. The body of the message up until the syntax error parses fine. Essentially, it's the same situation as:

```properties
key04 =
    This line is indented by 4 spaces,
junk here
```

test/fixtures/tab.ftl
```diff
@@ -0,0 +1,14 @@
+# Error (tab after =)
+key01 =	Value 01
```
Pike: I think this should be a success.

test/fixtures/tab.ftl
```diff
@@ -0,0 +1,14 @@
...
+
+# Error (tab before =)
+key02	= Value 02
```
Pike: This failing is right, as are the ones below.

test/fixtures/tab.ftl
stasm: Should the tab be trimmed from the value or kept verbatim?

stasm: I guess it makes sense to only trim syntax whitespace. I'd keep it.

stasm:
Thanks, @Pike. I made the tab character a `text_char` and also change the whitespace trimming logic to keep it as part of the value.

PR #168, opened 2018-08-22
Parse Comments line by line

stasm:
Spin-off from #160; the `Comment` part.

### DIFF
--- a/spec/fluent.ebnf
+++ b/spec/fluent.ebnf
@@ -1,15 +1,24 @@
 
-/* An FTL file defines a Resource. */
+/* An FTL file defines a Resource consisting of Entries. */
 Resource            ::= (Entry | blank_block | junk_line)*
+
+/* Entries are the main building blocks of Fluent. They define translations and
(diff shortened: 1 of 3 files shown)


stasm:
> Yeah, perhaps! I used BaseComment as a means to "tag" the output of the CommentLine parser. You could use PatternElement to do the same. Or, just pass the raw indent (not wrapped in any {type, value} object) into the code in abstract.mjs. The fact that it's not an AST node can also be used to discriminate it.

One more thing. The way I used `BaseComment` in the first comment, I didn't create an instance of it at any time. It's literally just for the purpose of matching the right `case` of a `switch`. I could have used any other value which is valid in a `case` comparison, like a string.

I'm not familiar with your exact use-case, but if you feel the need to temporarily store the indent as a node in the AST and then do something with it during a construction of another node, then we should look at the code and figure out if `PatternElement` is the right tool for it :)

In the second commit of this PR I dropped `BaseComment` in favor of a regular `Comment`.

PR #169, opened 2018-08-22
Parse Patterns line by line

stasm:
Spin-off from #160; the `Pattern` part.


stasm:
#171 is a better implementation of what I originally intended here. Closing.

PR #193, opened 2018-10-22
Don't normalize CRLF to LF in Junk

### LINKED ISSUE #184: Junk shouldn't normalize CRLF to LF.
The content of `Junk` should be the literal content which couldn't be parsed.

stasm:
Fix #184.

This doesn't change the parsing story for unsupported line endings (`\r`, `\u2028`, `\u2029`). Right now, that story is largely undefined.

### DIFF
--- a/syntax/grammar.mjs
+++ b/syntax/grammar.mjs
@@ -86,12 +86,14 @@ let CommentLine = defer(() =>
 
 /* ------------------------------------------------------------------------- */
 /* Adjacent junk_lines are joined into FTL.Junk during the AST construction. */
-let junk_line = defer(() =>
+let junk_line =
     sequence(
-        regex(/.*/),
-        line_end)
+        regex(/[^\n]*/),
+        either(
+            string("\u000A"),
+            eof()))
     .map(join)
-    .chain(into(FTL.Junk)));
+    .chain(into(FTL.Junk));
 
 /* --------------------------------- */
 /* Attributes of Messages and Terms. */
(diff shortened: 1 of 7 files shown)


syntax/grammar.mjs
```diff
@@ -86,12 +86,15 @@ let CommentLine = defer(() =>
...
 /* ------------------------------------------------------------------------- */
 /* Adjacent junk_lines are joined into FTL.Junk during the AST construction. */
-let junk_line = defer(() =>
+let junk_line =
     sequence(
         regex(/.*/),
-        line_end)
+        either(
+            string("\u000A"),
+            string("\u000D\u000A"),
```
stasm: I wonder if it would make sense to add other, technically unsupported, line endings here. In particular, I'm thinking of `\r`, `\u2028` and `\u2029`, which are not matched by `.` in the JavaScript regex implementation.

We don't need to include [all line terminators recognized by Unicode](https://en.wikipedia.org/wiki/Newline#Representation), but adding these extra three would make sure the reference parser can produce meaningful junk for files with these endings. Otherwise, the parser doesn't have any way to skip over the EOL in the `junk_line` production, and the parsing of the whole resource ends immediately.

Pike: Would `regex(/[^\n]*/`) be closer to our intent? And not change `line_end` at all?

stasm: This is a really good idea! Thanks.

I had some trouble making it work when I tested it. I was seeing additional `\n` where I was sure there shouldn't be any. And then I realized that I needed to use `echo -n` for my CLI testing...

stasm:
@Pike This is now ready for another round of review. Thanks!

PR #197, opened 2018-10-30
Dedent multiline text to preserve content indentation

### LINKED ISSUE #162: Preserve content indent in multiline Patterns
In Syntax 0.6 all leading indentation on each line of multiline Patterns is removed in the AST. The following Fluent code:

```properties
multi =
    Lorem ipsum dolor
      sit amet enim.
        Etiam ullamcorper.
```

parses as:

```json
{
    "type": "TextElement",
    "value": "Lorem ipsum dolor\nsit amet enim.\nEtiam ullamcorper.",
}
```

#### Problem statement

The indentation on each line conflates two orthogonal purposes:

  1. Fluent requires all text on a new line to be indented. Thanks to this indentation Fluent knows this is text content.
  2. The author of the example chose to indent line 2 and line 3 relative to line 1. The indentation is arbitrary and on top of what Fluent requires. It can be considered part of the _content_, just as the fact that this text is formatted over three separate lines is.

In Syntax 0.6, the only way for the localizer to have their indent preserved is to use `StringLiterals` with explicit whitespace in each line:

```properties
multi =
    Lorem ipsum dolor
    {"  "}sit amet enim.
    {"    "}Etiam ullamcorper.
```

This might be an acceptable solution to one-off situations, but it doesn't scale for multiline patterns, where `multi > 2`. One use-case for such patterns which require custom indentation is translations which define some multiline HTML to be used by a markup overlay mechanism.

#### Prior work

  - Python's [textwrap.dedent](https://docs.python.org/3/library/textwrap.html#textwrap.dedent). It is a runtime solution which isn't suitable to our needs of performing this work during the AST construction.
  - Scala's [stripMargin](https://www.scala-lang.org/files/archive/spec/2.12/01-lexical-syntax.html#multi-line-string-literals). It requires that a special character (by default `|`) be placed at the beginning of _each_ line of the string literal. I don't think this approach would be right for Fluent. In Fluent, typing text should be _the simplest_ thing to do. Consequently, text should have the sweetest syntax, i.e. no syntax at all. If anything, I'd consider using `|` for everything else except multiline text, but that's a different issue.
  - Swift's [closing quotation marks approach](https://docs.swift.org/swift-book/LanguageGuide/StringsAndCharacters.html#ID564). This takes advantage of the fact that string literals in Swift have delimiters. Again, in Fluent, the main form in which text exists (`TextElement`) doesn't have any syntax at all and a similar approach is not possible.

#### Proposal

I propose that the width of the indentation of the first line of a Pattern be considered the indent of the entire Pattern across all its lines. Any extra leading whitespace should be included in the AST as part of the Pattern.

In other words, the following example:
 
```properties
multi =
    Lorem ipsum dolor
      sit amet enim.
        Etiam ullamcorper.
```

will parse as:

```json
{
    "type": "TextElement",
    "value": "Lorem ipsum dolor\n  sit amet enim.\n    Etiam ullamcorper.",
}
```

In the rare cases when the first line of the Pattern needs to keep its indent, a `StringLiteral` should be used to 1) make it explicit and 2) define the indent required by the syntax. This is consistent with how creating Patterns with leading whitespace works right now. Without the explicit `StringLiteral`, all leading and trailing whitespace is trimmed during parsing.

```properties
multi =
    {"    "}Lorem ipsum dolor
      sit amet enim.
        Etiam ullamcorper.
```

parses as:

```json
{
    "type": "Placeable",
    "expression": {
        "type": "StringLiteral",
        "value": "    "
    }
},
{
    "type": "TextElement",
    "value": "Lorem ipsum dolor\n  sit amet enim.\n    Etiam ullamcorper.",
}
```

#### Requirements

For this to work, we'll need to forbid multiline Patterns which start on the same line as the identifier or the variant key.

```properties
now-forbidden = A Pattern which starts inline
    but continues as multiline. There's no indent
    on its first line whatsoever so it's not clear how
    much indent should be considered as required
    by the syntax and how much as indent to be
    kept in the AST.
```

```properties
above-snippet-fixed =
    A multiline Pattern must start on a new line.
    Now the syntax-required indentation is unambiguous
            and any extra leading whitespace will be kept
            in the AST.
```

stasm:
Fix #162.

### DIFF
--- a/syntax/abstract.mjs
+++ b/syntax/abstract.mjs
@@ -111,12 +111,16 @@ export function into(Type) {
     }
 }
 
+// Create a reducer suitable for joining adjacent nodes of the same type, if
+// type is one of types specified.
 function join_adjacent(...types) {
     return function(acc, cur) {
         let prev = acc[acc.length - 1];
         for (let Type of types) {
             if (prev instanceof Type && cur instanceof Type) {
-                join_of_type(Type, prev, cur);
+                // Replace prev with a new node of the same type whose value is
+                // the sum of prev and cur, and discard cur.
+                acc[acc.length - 1] = join_of_type(Type, prev, cur);
                 return acc;
             }
         }
@@ -124,20 +128,21 @@ function join_adjacent(...types) {
     };
 }
 
+// Join values of two or more nodes of the same type. Return a new node.
 function join_of_type(Type, ...elements) {
     // TODO Join annotations and spans.
     switch (Type) {
         case FTL.TextElement:
             return elements.reduce((a, b) =>
-                (a.value += b.value, a));
+                new Type(a.value + b.value));
         case FTL.Comment:
         case FTL.GroupComment:
         case FTL.ResourceComment:
             return elements.reduce((a, b) =>
-                (a.content += `\n${b.content}`, a));
+                new Type(a.content + "\n" + b.content));
         case FTL.Junk:
             return elements.reduce((a, b) =>
-                (a.content += b.content, a));
+                new Type(a.content + b.content));
     }
 }
 
@@ -154,9 +159,7 @@ function attach_comments(acc, cur) {
     }
 }
 
-const LEADING_BLANK_BLOCK = /^\n*/;
-const TRAILING_BLANK_INLINE = / *$/;
-
+// Remove the largest common indentation from a list of TextElements.
 function dedent(elements) {
     let indents = elements.filter(element => element instanceof FTL.Indent);
     let common = Math.min(...indents.map(indent => indent.value.length));
@@ -165,6 +168,9 @@ function dedent(elements) {
     }
     return elements;
 }
+
+const LEADING_BLANK_BLOCK = /^\n*/;
+const TRAILING_BLANK_INLINE = / *$/;
 
 function trim_text_at_extremes(element, index, array) {
     if (element instanceof FTL.TextElement) {
--- a/syntax/abstract.mjs
+++ b/syntax/abstract.mjs
@@ -48,7 +48,7 @@ export function list_into(Type) {
         case FTL.Pattern:
             return elements =>
                 always(new FTL.Pattern(
-                    elements
+                    dedent_whitespace(elements)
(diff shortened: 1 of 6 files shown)


stasm:
@Pike This is based on your branch from https://github.com/projectfluent/fluent/issues/162#issuecomment-415735496. Here's a summary of differences:

- Leading indent in block text is now preserved, too.
- Placeables on a new line with no indent count as indent of width 0. On your branch, they were skipped. I did this to make the "ruler rule" easier to visualize in all cases.
- I removed `FTL.Indent`. I don't have strong feeling about either way but I realized that it was the only non-base AST class which would never be used in an actual AST. I don't know if that's good or bad; we can figure it out later. Right now I went for a simple `typeof element === "string"` check instead.

syntax/abstract.mjs
```diff
@@ -154,18 +159,36 @@ function attach_comments(acc, cur) {
     }
 }
 
-const LEADING_BLANK = /^[ \n\r]*/;
-const TRAILING_BLANK = /[ \n\r]*$/;
+// Remove the largest common indentation from a list of PatternElements.
```
Pike: Extend this comment to make it expressive that all literal strings are indents, thanks to grammar.mjs separating those out?

syntax/abstract.mjs
```diff
@@ -111,33 +111,38 @@ export function into(Type) {
...
+// Create a reducer suitable for joining adjacent nodes of the same type, if
+// type is one of types specified.
 function join_adjacent(...types) {
     return function(acc, cur) {
         let prev = acc[acc.length - 1];
         for (let Type of types) {
             if (prev instanceof Type && cur instanceof Type) {
-                join_of_type(Type, prev, cur);
+                // Replace prev with a new node of the same type whose value is
+                // the sum of prev and cur, and discard cur.
```
Pike: ... `discard cur and prev.`?

stasm: Do you think it's needed given that the comment starts with `Replace...`?

Pike: I could go without the `, and discard ...` altogether, too.

PR #199, opened 2018-10-30
Forbid control characters and all non-characters

### LINKED ISSUE #182: Forbid C1 control chars in any_char
Follow-up to #174. Right now, `regular_char` is defined as:

```
/* Any Unicode character excluding C0 control characters (but including tab),
 * surrogate blocks and non-characters (U+FFFE, U+FFFF).
 * Cf. https://www.w3.org/TR/REC-xml/#NT-Char */
any_char            ::= [\\u{9}\\u{20}-\\u{D7FF}\\u{E000}-\\u{FFFD}]
                      | [\\u{10000}-\\u{10FFFF}]
```

The `U+21` - `U+D7FF` range includes more control characters which make little sense in translations:

> The Unicode control characters cover U+0000—U+001F (C0 controls), U+007F (delete), and U+0080—U+009F (C1 controls). Unicode only specifies semantics for U+001C—U+001F, U+0009—U+000D, and U+0085. The rest of the control characters are transparent to Unicode and their meanings are left to higher-level protocols. <sup>[source](https://en.wikipedia.org/wiki/C0_and_C1_control_codes#Encoding_interoperability)</sup>

I'd like to suggest that we forbid characters from `U+7F` to `U+9F`, inclusive.  This would be inline with the recommendation included in the definition of [`NT-Char`](https://www.w3.org/TR/REC-xml/#NT-Char) in the XML spec, although with the exception of the `U+85` (`NEL`, `Next Line`) character. `NEL` is equivalent to `CRLF` and appears to be mostly used on [IBM mainframes running z/OS](https://www.w3.org/TR/newline). I think it's OK to forbid it in Fluent (we can always revisit later if need be).

stasm:
Fix #182.

### DIFF
--- a/syntax/grammar.mjs
+++ b/syntax/grammar.mjs
@@ -376,14 +376,45 @@ let Function =
 /* Content Characters
  *
  * Translation content can be written using most Unicode characters, with the
- * exception of C0 control characters (but allowing tab), surrogate blocks and
- * non-characters (U+FFFE, U+FFFF).
+ * exception of C0 and C1 control characters (but including TAB), DEL,
+ * surrogate blocks and non-characters. New lines are excluded because they are
+ * forbidden in quoted_text and handled with additional logic in block_text.
  */
 
 let any_char =
     either(
-        charset("\\u{9}\\u{20}-\\u{D7FF}\\u{E000}-\\u{FFFD}"),
-        charset("\\u{10000}-\\u{10FFFF}"));
+        // TAB
+        charset("\\u{9}"),
+        // Basic Latin.
+        charset("\\u{20}-\\u{7E}"),
+        // Skip DEL and C1 controls.
+        // Resume at Latin-1.
+        charset("\\u{A0}-\\u{D7FF}"),
+        // Skip surrogate codepoints.
+        // Resume at the U+E000 PUA.
+        charset("\\u{E000}-\\u{FDCF}"),
+        // Skip the non-character block U+FDD0..FDEF */
+        // Resume at U+FDF0.
+        charset("\\u{FDF0}-\\u{FFFD}"),
+        // Skip U+FFFE and U+FFFF.
+        // Resume at astral planes. The last two codepoints of each plane are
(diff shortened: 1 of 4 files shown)


syntax/grammar.mjs
```diff
@@ -376,14 +376,39 @@ let Function =
...
+        charset("\\u{20}-\\u{7E}"),
+        // Skip DEL and C1 controls.
+        // Resume at Latin-1.
+        charset("\\u{A0}-\\u{D7FF}"),
+        // Skip surrogate codepoints.
+        // Resume at the U+E000 PUA.
+        charset("\\u{E000}-\\u{FDCF}"),
+        // Skip the non-character block U+FDD0..FDEF */
+        // Resume at U+FDF0.
+        charset("\\u{FDF0}-\\u{FFFD}"),
```
stasm: I tried adding a test with `abc<U+D800>def` (the first high surrogate) and somehow, this `charset` parser successfully parses it. I'm not sure what's going on yet.

stasm: OK, I think I figured out why this was so. In UTF-8, surrogate pair as well as lonely surrogates are ill-formed. Conformant decoders must handle it in some way:

> When faced with this illegal byte sequence while transforming or interpreting, a UTF-8 conformant process must treat the first byte [of the example] as an illegal termination error: for example, either signaling an error, filtering the byte out, or representing the byte with a marker such as FFFD (REPLACEMENT CHARACTER). <sub>[source](http://unicode.org/faq/utf_bom.html#gen8)</sub>

It looks like node defaults to using the `U+FFFD` (�) character in such cases.  And that's exaclt the character the parser sees here.

To write a test for rejecting surrogates in the parser, we'd need to have a different test runner which would build a test string in JavaScript, rather than reading it from a file using the UTF-8 encoding.

Pike: I think we should rely on UTF-8 protecting us here.

stasm:
Closing this in favor of #207.

Issue #194, opened 2018-10-22
Unicode Escapes should cover more unicode planes


Pike:
Now that we have more than the basic Unicode plane in the Fluent syntax, we should also support them in the Unicode escapes.

I suggest to use 4 or 6 digits, based on earlier conversations.

I wonder if we should exclude surrogate pairs at the same time, to prevent `\uD83D\uDE02` in favor of `\u01F602`? The UTF-16 encoding these imply feel very implementation dependent to me.

stasm:
Are there characters which are better written as escapes rather than actual glyphs?

> I suggest to use 4 or 6 digits, based on earlier conversations.

I'd like to throw the `\u{…}` proposal into the mix, too. The number of hexdigits between the braces can be between 1 and 6. Examples: `\u{9}`, `\u{A0}`, `\u{1F602}`.

Pike:
logical operators, pff. Support `4`, `6`. Not support `5`.

Pike:
> Would `"\u00a0ff"` be interpreted as `<nbsp>ff` or as `ꃿ`?

Yeah, that's a problem. Maybe just `\U00a0ff` ?
```
unicode_escape      ::= "\\u" /[0-9a-fA-F]{4}/
        | "\\U" /[0-9a-fA-F]{6}/
```

I'm not a fan of `\u{}`, for one because it gives `{}` a different meaning in that context. I'm also concerned about the amount of work we'd have to throw at it.

zbraniecki:
@manishearth - do you have any thoughts on this from Rust? In particular, should we go for 6 digits, or 8?

Manishearth:
Rust allows for all code points in source files too, the reason escapes exist is to let people specify them explicitly, especially in cases where there are invisible code points.

You can also do something like `\u[..]`, pick a brace syntax

stasm:
To summarize: We could either have two syntaxes:

```properties
terms-u = Terms{"\u00A0"}and{"\u00A0"}Conditions
terms-U = Terms{"\U0000A0"}and{"\U0000A0"}Conditions
```

Or a single one using some kind of delimiters:

```properties
terms-brace = Terms{"\u{A0}"}and{"\u{A0}"}Conditions
terms-bracket = Terms{"\u[A0]"}and{"\u[A0]"}Conditions
terms-paren = Terms{"\u(A0)"}and{"\u(A0)"}Conditions
terms-angle = Terms{"\u<A0>"}and{"\u<A0>"}Conditions
```

[...]

PR #203, opened 2018-11-06
Store unescaped content in StringLiteral.value and raw content in StringLiteral.raw

### LINKED ISSUE #195: Create explicit AST node for escape sequences
Right now, the runtime behavior of `\XXX` is only implicitly defined by the js runtime.

To fix this, we should have a dedicated object in our AST tree that represents a escaped character, that is Unicode escapes, and escapes for syntax chars.

This is half of #156 with a different rationale. The AST might look similar to https://github.com/projectfluent/fluent/issues/123#issuecomment-431303286.

stasm:
This also forbids escape sequences representing surrogate code points, following @Pike's suggestion from https://github.com/projectfluent/fluent/issues/194#issue-372533250. If we allow them, then their handling becomes environment/implementation dependent. E.g. they are ill-formed in the UTF-8 encoding.

### DIFF
--- a/spec/CHANGELOG.md
+++ b/spec/CHANGELOG.md
@@ -86,6 +86,14 @@
     present in `TextElements`. `{"}"}` can be used to insert a literal
     closing brace.
 
+  - Store both the raw and the unescaped value in `StringLiteral`. (#203)
+
+    `StringLiteral.value` has been change to store the unescaped ("cooked")
+    value of the string literal: all known escape sequences are replaced with
+    the characters they represent. `StringLiteral.raw` has been added and
+    stores the raw value as it was typed by the author of the string literal:
+    escapes sequences are not processed in any way.
+
   - Don't normalize line endings in `Junk`. (#184)
 
     Junk represents a literal slice of unparsed content and shouldn't have
--- a/syntax/abstract.mjs
+++ b/syntax/abstract.mjs
@@ -115,6 +115,9 @@ export function into(Type) {
                 }
                 return always(new Type(expression));
             };
+        case FTL.StringLiteral:
+            return raw =>
+                always(new Type(raw, unescape(raw)));
         default:
             return (...args) =>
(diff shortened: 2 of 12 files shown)


stasm:
> That said, maybe that's OK within this context, but we'll need comments for that.

Where would you put it?

> Also, please keep the current semantics of .value, as most tools use that and will want to continue to use that.

Pike:
> Where would you put it?

In `ast.mjs`, so that consumers of that API know what to use when.

Pike:
This is technically OK, but I disagree with the choice out of my own experience in compare-locales.

Practically, this is going to add a significant risk in particular to the Pontoon support, as we're changing the data the existing `.value` API returns.

Not going to block this, but also not going to r+ this, so removing myself from the review.

Pike:
Meh, can't find a way to actually remove myself without showing my previous r-.

stasm:
> Practically, this is going to add a significant risk in particular to the Pontoon support, as we're changing the data the existing .value API returns.

Thank you for you comment. I think I'm misunderstanding something here. When we update Pontoon to Syntax 0.8, we'll change the code which deals with `StringLiterals` to use the `.raw` field rather than `.value`. It sounds like this will solve this problem completely. We'll need to make changes to tools anyways. Is there more to this which I'm not seeing?

stasm:
Actually, it looks like updating the serializer to use `.raw` for `StringLiterals` (which we'll need to do anyways) will be enough for Pontoon: https://github.com/mozilla/pontoon/blob/d55d260554d13173e4ee39629becbd0ab3a264af/pontoon/base/static/js/fluent_interface.js#L435.

Pike:
My point is that I made the same exact API decision, and it's been a source of bugs (just this week, actually).

That said, I don't want to spend more time on this ticket.

stasm:
OK, thanks again. I'm going to merge this as-is.

Issue #189, opened 2018-10-19
Consider accessing variants of term attributes


stasm:
Right now, the `-term.attr[name]` syntax produces an error. Keeping in mind that term attributes are meant to describe their grammatical properties, are there use-cases for multiple variants of them?

@Pike pointed me to [Vowel harmony](https://en.wikipedia.org/wiki/Vowel_harmony). I also recall a conversation about [Welsh mutations](https://en.wikibooks.org/wiki/Welsh/Mutations) which might be relevant here.

stasm:
The parameterized terms proposal (#176), which supersedes variant lists, doesn't solve this issue _per se_. We still need to introduce new syntax to access variants of attributes defined on terms. The two candidates are:

1. `-term.attr(arg: "value")`
2. `-term(arg: "value").attr`

I think both approaches have their merits and can be good choices depending on the use-case. Perhaps we should all both?

First, let's consider a term defined like the following (see [The Chicago Manual of Style](https://books.google.pl/books?id=EBMlLYwqmjYC&q="When+a+pronoun+is+used+to+refer+to+a+vessel")):

```properties
-ship = {$first-letter ->
   *[lower] ship
    [upper] Ship
}
.gender = {$style ->
   *[traditional] feminine
    [chicago] neuter
}
```

My preference here goes to the first proposed syntax: `-ship.gender(style: "chicago")`. Clearly the `style` argument is related to the attribute itself. It makes sense to me that the call expression comes after the attribute accessor. We are parameterizing the attribute independently of the term value.

OTOH, let's now consider a term defined like the following:

```properties
# A contrived example in English.
-echo = {$count->
   *[one] Echo
    [other] Echos
}
.last-letter = {$count->
   *[one] o
    [other] s
}
```

Here, both syntaxes are reasonable:

1. `-echo.last-letter(count: 2)`
2. `-echo(count: 2).last-letter`

I think I'd prefer to use syntax no. 2 in this case. Semantically, the count is a property of the value _of the term_, rather than the value of the attribute, like in the previous example.

stasm:
The first syntax proposed above makes it easier to implement macros in the runtime:

    -ship.gender(style: "chicago")
    -echo.last-letter(count: 2)

The `AttributeExpression` can be first resolved to an `Attribute`. Then, the `CallExpression` is used to get the desired variant from the `Attribute`'s value.

The second proposed syntax would require a closure-like resolution semantics.

    -echo(count: 2).last-letter

[...]

Issue #80, opened 2018-01-15
Dynamic message references


stasm:
It is sometimes desired to parametrize message references in placeables. In this issue I'd like to propose a new argument type, extending `FluentType` which could be used to programmatically pass message references as arguments to messages.

#### Problem Statement

Redundancy is considered good for localization. It allows localizers to tailor the wording and the grammar of the translation of each particular case. Also see [Fluent Good Practices](https://github.com/projectfluent/fluent/wiki/Good-Practices-for-Developers).

In general, the pattern of having one message per item is preferred over factoring the action out to its own message (`Delete This { $item }`) and passing the translated `item` in some way.

```properties
# Having two separate messages allows localizers
# to customize translations in each, if needed.
delete-picture = Delete This Picture
delete-video = Delete This Video
```

In some cases, however, this pattern doesn't scale well.

Consider this example from Firefox ([source](https://hg.mozilla.org/l10n/gecko-strings/file/6ed6a4050107/browser/chrome/browser/browser.properties#l769)):

```properties
# %S is the website origin (e.g. www.mozilla.org)
getUserMedia.sharingMenuCamera = %S (camera)
getUserMedia.sharingMenuMicrophone = %S (microphone)
getUserMedia.sharingMenuAudioCapture = %S (tab audio)
getUserMedia.sharingMenuApplication = %S (application)
getUserMedia.sharingMenuScreen = %S (screen)
getUserMedia.sharingMenuWindow = %S (window)
getUserMedia.sharingMenuBrowser = %S (tab)
getUserMedia.sharingMenuCameraMicrophone = %S (camera and microphone)
getUserMedia.sharingMenuCameraMicrophoneApplication = %S (camera, microphone and application)
getUserMedia.sharingMenuCameraMicrophoneScreen = %S (camera, microphone and screen)
getUserMedia.sharingMenuCameraMicrophoneWindow = %S (camera, microphone and window)
getUserMedia.sharingMenuCameraMicrophoneBrowser = %S (camera, microphone and tab)
getUserMedia.sharingMenuCameraAudioCapture = %S (camera and tab audio)
getUserMedia.sharingMenuCameraAudioCaptureApplication = %S (camera, tab audio and application)
getUserMedia.sharingMenuCameraAudioCaptureScreen = %S (camera, tab audio and screen)
getUserMedia.sharingMenuCameraAudioCaptureWindow = %S (camera, tab audio and window)
[...]

PR #205, opened 2018-11-06
Add Parameterized Terms

### LINKED ISSUE #176: Parameterized Terms
_a.k.a. The Return of Macros._

In #102 we added `VariantLists`. They are a solution to abusing `SelectExpressions` by not defining the `selector`. Variants of a `VariantList` can be retrieved externally from where the term is referenced with a `VariantExpression`.

The design of `VariantLists` has its own shortcomings which I outlined in https://github.com/projectfluent/fluent/issues/102#issuecomment-385069390 and which required the addition of specialized [code in `abstract.mjs`](
https://github.com/projectfluent/fluent/blob/20a5d258504b3bf8029224bf9c213d465b90af7b/syntax/abstract.mjs#L77-L83) to tightly control where exactly `VariantLists` are allowed.

At the time, we accepted these shortcomings as trade-offs and also because we didn't find other alternatives. I think I may have stumbled across an idea which is worth bringing it up here.

#### Proposal

Rather than treating terms with `VariantLists` for value as dictionaries whose members must be accessed from the outside, what if we encapsulated the member access in a form of a function?

In other words, rather than defining the following term as a `VariantList`:

```properties
# Before, with VariantLists
-fxaccount-brand-name = {
    [lowercase] account Firefox
   *[uppercase] Account Firefox
}

sync-signedout-account-title = Connetti il tuo { -fxaccount-brand-name[lowercase] }
```

We could parameterize it with the `$case` local variable and use a regular `Pattern` as the value:

```properties
# After, with parameterized terms
-fxaccount-brand-name($case) = { $case ->
    [lowercase] account Firefox
   *[uppercase] Account Firefox
}

# Call it similar to a function
sync-signedout-account-title = Connetti il tuo { -fxaccount-brand-name("lowercase") }
```

Using the default variant, or referencing a term which doesn't have any parameters would use the current `TermReference` syntax without any paranthesis:

```properties
# English FTL
sync-signedout-account-title = Connect to your { -fxaccount-brand-name }
```

#### Advantages

- Values of terms can always be `Patterns` again.
- Accessing a specific variant uses the `CallExpression` syntax. No need for the specialized `VariantExpression`.
- Accessing variants would now use the same semantics as `SelectExpressions` in `Patterns` do right now. New features of `SelectExpressions`, like #4, would automatically be available in terms.
- Local variables can be used inside of the term's value, too. The above example doesn't show case this, but you could imagine a different term using a `$number` variable. In the current design, the name of this variable is effectively hardcoded in the term and all messages referencing it must have the variable of this exact name passed into them by the app. This would limit the utility of private terms, which we might introduce in the future.

#### Open questions

1. How should the scoping rules work exactly? Should terms be allowed to access variables passed to messages which reference them, or should all variables be passed to them explicitly?
2. Should attributes use the parameters defined on the term id, or should there be a way to define (possibly different) parameters on the attribute id?
3. Related, when referencing attributes, should the parameters be attached to the term (`{-brand-name("lowercase").gender}`) or to the attribute (`{-brand-name.gender("lowercase")}`)?
4. Could we use named arguments to define default values of parameters?

stasm:
Fix #176.

### DIFF
--- a/test/fixtures/term_parameters.ftl
+++ b/test/fixtures/term_parameters.ftl
@@ -1,0 +1,8 @@
+-term = { $arg ->
+   *[key] Value
+}
+
+key01 = { -term }
+key02 = { -term() }
+key03 = { -term(arg: 1) }
+key04 = { -term("positional", narg1: 1, narg2: 2) }
--- a/spec/CHANGELOG.md
+++ b/spec/CHANGELOG.md
@@ -58,6 +58,59 @@
             on the second line of its value.
     ```
 
+  - Introduce parameterized `Terms`. (#176)
+
+    References to `Terms` can now receive parameters which will be used by
+    the runtime as values of variables referenced from within the `Term`.
+    This allows `Terms` to use regular `Patterns` as values, rather than
+    `VariantLists`:
+
+    ```properties
+    # A Term with a VariantList as a value.
+    -thing = {
+       *[definite] the thing
+       *[indefinite] a thing
+    }
+
+    this = This is { -term[indefinite] }.
+    ```
+
+    ```properties
+    # A parametrized Term with a Pattern as a value.
+    -thing = { $article ->
+       *[definite] the thing
+       *[indefinite] a thing
+    }
+
+    this = This is { -thing(article: "indefinite") }.
+    ```
+
+    Since `Patterns` can be nested, this feature allows more complex
+    hierarchies of term values:
+
+    ```properties
+    # A parametrized Term with nested Patterns.
+    -thing = { $article ->
+       *[definite] { $first-letter ->
+           *[lower] the thing
+            [upper] The thing
+        }
+        [indefinite] { $first-letter ->
+           *[lower] a thing
+            [upper] A thing
+        }
+    }
+
+    this = This is { -term(first-letter: "lower", article: "indefinite") }.
+    ```
+
+    Parameters must be named; positional parameters are ignored. If a
+    parameter is omitted then the regular default variant logic applies. The
+    above example could thus be written as `{-term(article: "indefinite")}`
+    and the `lower` variant would be used because it is marked as the default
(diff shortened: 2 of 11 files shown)


stasm:
> Can you please open a separate issue for the changes to function names?

Sure: #209.

> I'm wondering if we could formulate the parametrized terms better. How about replacing TermReference in InlineExpression with TermExpression, and maybe factor the argument list a bit to include the full call expression postfix?

That's an interesting approach. However, I see `CallExpression` as being useful for macros because of its generality. We might want to enable the `-term.attr()` syntax soon (#189). And later, dynamic references will likely require `$var()` (#80).

> I picture this from a resolver implementation, and I expect the implementation to set up the environment for Terms to be rather different than to that of a call into the globals. So implementers might do the above anyway, but they'd need to piece that together?

Or they can check the type of the `callee`.

test/fixtures/term_parameters.ftl
```diff
@@ -0,0 +1,8 @@
+-term = { $arg ->
+   *[key] Value
+}
+
+key01 = { -term }
+key02 = { -term() }
+key03 = { -term(arg: 1) }
+key04 = { -term("positional", narg1: 1, narg2: 2) }
```
Pike: Are positional arguments intended to be legal?

stasm: They're not used by the runtime to resolve the term so even if they're passed, they're just ignored. We could forbid them in `abstract` for now, in case we find a use for them in the future. How does that sound to you?

Pike: I wondered about `abstact`, too, but no idea if there's going to be a use or not. I'm not a friend of "for now", either, so, well. Yeah, I filed a documentation bug. No strong opinion.

stasm: Let's leave them as valid, then. Thanks for filing the issue about the docs.

PR #207, opened 2018-11-06
 Allow all Unicode characters

### LINKED ISSUE #182: Forbid C1 control chars in any_char

stasm:
Fix #182, but differently. An alternative to #199.

### DIFF
--- a/syntax/grammar.mjs
+++ b/syntax/grammar.mjs
@@ -392,15 +392,13 @@ let Identifier =
 /* -------------------------------------------------------------------------- */
 /* Content Characters
  *
- * Translation content can be written using most Unicode characters, with the
- * exception of C0 control characters (but allowing tab), surrogate blocks and
- * non-characters (U+FFFE, U+FFFF).
+ * Translation content can be written using any Unicode characters. However,
+ * some characters are considered special depending on the type of content
+ * they're in. See text_char and quoted_char for more information.
  */
 
 let any_char =
-    either(
-        charset("\\u{9}\\u{20}-\\u{D7FF}\\u{E000}-\\u{FFFD}"),
-        charset("\\u{10000}-\\u{10FFFF}"));
+    charset("\\u{0}-\\u{10FFFF}");
 
 /* -------------------------------------------------------------------------- */
 /* Text elements
@@ -418,10 +416,11 @@ let special_text_char =
         string("{"),
         string("}"));
 
-let text_char =
+let text_char = defer(() =>
     and(
+        not(line_end),
         not(special_text_char),
-        any_char);
+        any_char));
 
 let indented_char =
     and(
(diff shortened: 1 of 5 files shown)


test/fixtures/any_char.ftl
```diff
@@ -0,0 +1,8 @@
+# ERROR       ↓ BEL, U+0007
```
Pike: Adjust the comments such that these aren't errors anymore? All of them in this file.

syntax/grammar.mjs
```diff
@@ -390,15 +390,13 @@ let Identifier =
 /* -------------------------------------------------------------------------- */
 /* Content Characters
  *
- * Translation content can be written using most Unicode characters, with the
- * exception of C0 control characters (but allowing tab), surrogate blocks and
- * non-characters (U+FFFE, U+FFFF).
+ * Translation content can be written using any Unicode characters. However,
+ * some characters are considered special depending on the type of content
+ * they're in. See text_char and quoted_char for more information.
```
Pike: Add a reference to recommendations.md here, too? Similar to how we link to valid.md in this file?

test/fixtures/any_char.ftl
stasm: Oops, thanks :)

PR #211, opened 2018-11-07
 Stop parsing Junk on lines which look like Entries 

### LINKED ISSUE #185: Adjacent broken entries shouldn't be merged into a single Junk
```properties
foo = {1foo}
bar = {2bar}
```

should produce two Junk entries in the AST. Right now they're are merged into a single one because junk is first parsed line-by-line (`junk_line`) and it's only combined in `abstract.mjs`. <del>We'll likely need two grammar productions, `junk_first_line` and `junk_follow_line`, so that we can distinguish between them in `abstract.mjs` and only join the first line with its follow lines.</del>

stasm:
Fix #185.

The second commit has a few more tests for this specific change. I didn't want them to get lost in the many changes to fixtures this PR has. I'll squash it when landing.

### DIFF
--- a/spec/fluent.ebnf
+++ b/spec/fluent.ebnf
@@ -1,6 +1,6 @@
 
 /* An FTL file defines a Resource consisting of Entries. */
-Resource            ::= (Entry | blank_block | junk_line)*
+Resource            ::= (Entry | blank_block | junk)*
 
 /* Entries are the main building blocks of Fluent. They define translations and
  * contextual and semantic information about the translations. During the AST
@@ -18,7 +18,13 @@ Term                ::= "-" Identifier blank_inline? "=" blank_inline? Value Att
  * the AST construction. */
 CommentLine         ::= ("###" | "##" | "#") ("\u0020" /.*/)? line_end
 
-/* Adjacent junk_lines are joined into FTL.Junk during the AST construction. */
+/* Junk represents unparsed content.
+ *
+ * Junk is parsed line-by-line until a line is found which looks like it might
+ * be a beginning of a new message, term, or a comment. Any whitespace
+ * following a broken Entry is also considered part of Junk.
+ */
+junk                ::= junk_line (junk_line - "#" - "-" - [a-zA-Z])*
 junk_line           ::= /[^\n]*/ ("\u000A" | EOF)
 
 /* Attributes of Messages and Terms. */
--- a/spec/fluent.ebnf
+++ b/spec/fluent.ebnf
@@ -1,6 +1,6 @@
 
 /* An FTL file defines a Resource consisting of Entries. */
-Resource            ::= (Entry | blank_block | junk)*
+Resource            ::= (Entry | blank_block | Junk)*
 
 /* Entries are the main building blocks of Fluent. They define translations and
  * contextual and semantic information about the translations. During the AST
@@ -24,7 +24,7 @@ CommentLine         ::= ("###" | "##" | "#") ("\u0020" /.*/)? line_end
  * be a beginning of a new message, term, or a comment. Any whitespace
  * following a broken Entry is also considered part of Junk.
  */
-junk                ::= junk_line (junk_line - "#" - "-" - [a-zA-Z])*
+Junk                ::= junk_line (junk_line - "#" - "-" - [a-zA-Z])*
 junk_line           ::= /[^\n]*/ ("\u000A" | EOF)
 
 /* Attributes of Messages and Terms. */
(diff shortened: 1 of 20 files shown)


test/fixtures/call_expressions.json
```diff
@@ -213,7 +213,7 @@
         {
             "type": "Junk",
             "annotations": [],
-            "content": "shuffled-args = {FUN(1, x: 1, \"a\", y: \"Y\", msg)}\n"
+            "content": "shuffled-args = {FUN(1, x: 1, \"a\", y: \"Y\", msg)}\n\n"
```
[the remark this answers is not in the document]
stasm: Great question! I'm not happy with the answer, but the good news is that this PR will help in making this right.

The reference fixtures in `fluent-syntax` are copied directly from the reference parser's tests. They're not generated by the tooling parser like the other kinds of fixtures. I've taken a note to document this in a `README`.

So what you're seeing in `fluent-syntax/test/fixtures_reference` are actually the [reference parser's fixtures from Syntax 0.7](https://github.com/projectfluent/fluent/blob/v0.7.0/test/fixtures/call_expressions.json#L216).

The actual output of the tooling parser includes those trailing blank lines. You can verify that in the Playground. The `fixtures_reference` tests pass in `fluent-syntax` because the [test runner explicitly ignores `Junk`](https://github.com/projectfluent/fluent.js/blob/a6bab32256e0148aa3256ba949cde1b583be0fe3/fluent-syntax/test/reference_test.js#L49-L52) due to its being parsed differently in Syntax 0.7. With this PR, we're getting much closer to being able to test junk too :)

spec/fluent.ebnf
```diff
@@ -18,7 +18,13 @@ Term                ::= "-" Identifier blank_inline? "=" blank_inline? Value Att
...
 CommentLine         ::= ("###" | "##" | "#") ("\u0020" /.*/)? line_end
 
-/* Adjacent junk_lines are joined into FTL.Junk during the AST construction. */
+/* Junk represents unparsed content.
+ *
+ * Junk is parsed line-by-line until a line is found which looks like it might
+ * be a beginning of a new message, term, or a comment. Any whitespace
+ * following a broken Entry is also considered part of Junk.
+ */
+junk                ::= junk_line (junk_line - "#" - "-" - [a-zA-Z])*
```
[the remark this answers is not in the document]
stasm: Good catch. I'll try to clean up the naming conventions, probably after 0.8.

spec/fluent.ebnf
```diff
@@ -18,7 +18,13 @@ Term                ::= "-" Identifier blank_inline? "=" blank_inline? Value Att
...
 CommentLine         ::= ("###" | "##" | "#") ("\u0020" /.*/)? line_end
 
-/* Adjacent junk_lines are joined into FTL.Junk during the AST construction. */
+/* Junk represents unparsed content.
+ *
+ * Junk is parsed line-by-line until a line is found which looks like it might
+ * be a beginning of a new message, term, or a comment. Any whitespace
+ * following a broken Entry is also considered part of Junk.
+ */
+Junk                ::= junk_line (junk_line - "#" - "-" - [a-zA-Z])*
```
Pike: This rendering to ebnf doesn't make any sense, right?

stasm: I think it does. We're using the EBNF syntax as defined in the [XML spec](https://www.w3.org/TR/xml/#sec-notation), with an extension of allowing regexes in a few places. The XML one reads:

     A - B
        matches any string that matches A but does not match B.

So `junk_line - "#"` matches a line of junk which doesn't start with a `#`. I think that's exactly what we want to say here.

stasm: > # foo\n does match junkline, but it doesn't match #, so it's a junk line.

Can you rephrase this please?

To me, the EBNF in this PR clearly expresses the intent. This is already a slippery slope because we're trying to define how to parse unparsed content. I don't want to overthink it. I'm also not sure how you'd like to write this differently. I could look at a PR if you'd like to prepare one :)

Pike: I just took the literal quote from the xml spec, and replaced `A` and `B` with `junk_line` and `#`, resp. And tested that against a candidate line line `# foo\n`.

I know that you don't believe in the value of the EBNF, and I don't care much either about this one.

stasm: > I just took the literal quote from the xml spec, and replaced A and B with junk_line and #, resp. And tested that against a candidate line line # foo\n.

Ah, I see what you mean, thanks. `#` matches `# foo\n` partially and that's enough for a negative lookahead to work here. I guess we could try to refactor this into something like `sequence(and(not("#"), any_char), junk_line)` but it would require special handling of blank lines inside of junk. (`any_char` doesn't parse newlines.) All in all, I favor the expressiveness of the approach I implemented in this PR.

PR #212, opened 2018-11-08
Add parameterized Attributes

### LINKED ISSUE #189: Consider accessing variants of term attributes

stasm:
Fix #189. Add syntax to allow expressions of the form `-term.attr(arg: "value")`, e.g.:

    -ship.gender(style: "chicago")
    -echo.last-letter(count: 2)


stasm:
This depends on #217 which is the first commit here. The second commit has the changes to the code, and the third one—tests. We were missing some coverage when it comes to validating expressions in different contexts.

syntax/abstract.mjs
```diff
@@ -10,6 +10,16 @@ import {always, never} from "../lib/combinators.mjs";
...
+        case FTL.AttributeExpression:
+            return ([ref, name]) => {
+                let ref_is_valid =
+                    ref instanceof FTL.MessageReference
+                    || ref instanceof FTL.TermReference;
+                if (ref_is_valid) {
+                    return always(new Type(ref, name));
+                }
+                return never(`Invalid ref type: ${ref.type}.`);
+            };
```
Pike: I think this part is not needed with the reduced changes after the rebase? Attributes are always on messages or terms still.

stasm: You're right, but I'd like to keep this here for the sake of completeness. I think it complements the validation of `CallExpression`. Let's keep it for now and I'll see if I can remove it in a follow-up.

PR #217, opened 2018-11-14
Make Messages and Functions be in the same namespace

stasm:
Forbid all uppercase ASCII identifiers for `Messages` and effectively make Messages and Functions co-exist in the same namespace.

```properties
## BEFORE

# FUNC is a FunctionReference
one = {FUNC()}
# FUNC is a MessageReference
two = {FUNC.attr}
```

With a single namespace for both Messages and Functions, `two` becomes a syntax error.

```properties
## AFTER

# FUNC is a FunctionReference
one = {FUNC()}
# FUNC is still a FunctionReference → Syntax Error
# (FunctionReference is not a valid `ref` of an `AttributeExpression`.)
two = {FUNC.attr}
```

### DIFF
--- a/spec/fluent.ebnf
+++ b/spec/fluent.ebnf
@@ -55,31 +55,33 @@ block_placeable     ::= blank_block blank_inline? inline_placeable
  * syntax/abstract.mjs. */
 InlineExpression    ::= StringLiteral
                       | NumberLiteral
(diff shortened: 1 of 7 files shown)


stasm:
Yep, this is a wontfix.

PR #219, opened 2018-11-16
Change line endings in comments to line_end

stasm:
The `/.*/` regex doesn't match any of the line endings recognized by the JS RegExp engine, i.e. `\n`, `\r`, `\u2028`, `\u2029`. These are different from the line endings supported by the rest of the Fluent Syntax. This change unifies the line endings in all grammar productions.


stasm:
> We should do the same for comments and junk. Not sure if they should share a rest_of_line production, but they should implement the same solution for the problem.

I think they are different and they shouldn't share the same solution. Line endings in Comments are normalized to LF. Line endings in Junk are unchanged, as per https://github.com/projectfluent/fluent/issues/184.

PR #221, opened 2018-11-29
 Store attribute access and call args on reference expressions 

stasm:
Right now, we store complex member expressions as nested trees of AST nodes. As an example, a parameterized term attribute of the form `-term.attr(arg: "value")` parses as:

    CallExpression {
        callee: AttributeExpression {
            ref: TermReference {
                id: Identifier {name: "term"}
            },
            name: Identifier {name: "attr"}
        },
        positional: [],
        named: [NamedArgument {...}]
    }

This is a well established pattern in general-purpose languages. It is capable of handling deeply nested hierarchies of expressions. In Fluent, however, we only ever go so deep. All the possible nestings are known up-front and are defined by the grammar.

Furthermore, the above AST structure makes tooling harder. Checking for a `TermReference` means checking for four different AST structures (corresponding to `-term`, `-term()`, `-term.attr`, and `-term.attr()`).

In this PR, I've tried a different approach: the attribute access and the call arguments are stored on the reference node itself:

    TermReference {
        id: Identifier {name: "term"},
        attribute: Identifier {name: "attr"}
        args: CallArguments {
            positional: [],
            named: [NamedArgument {...}]
        }
    }

I implemented this approach for `MessageReference`, `TermReference`, `VariableReference` and `FunctionReference`.

### DIFF
--- a/spec/fluent.ebnf
+++ b/spec/fluent.ebnf
@@ -52,11 +52,7 @@ block_placeable     ::= blank_block blank_inline? inline_placeable
  * syntax/abstract.mjs. */
 InlineExpression    ::= StringLiteral
                       | NumberLiteral
-                      | CallExpression
-                      | AttributeExpression
-                      | MessageReference
-                      | TermReference
-                      | VariableReference
+                      | ReferenceExpression
                       | inline_placeable
 
 /* Literals */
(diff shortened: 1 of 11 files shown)


syntax/ast.mjs
```diff
@@ -192,6 +177,15 @@ export class Variant extends SyntaxNode {
     }
 }
 
+export class CallArguments extends SyntaxNode {
+    constructor(positional = [], named = []) {
+        super();
+        this.type = "CallArguments";
+        this.positional = positional;
+        this.named = named;
```
stasm: `CallArguments` is a separate `SyntaxNode` for two reasons:

1. The tooling parser can store the span of `(...)` on it.
2. In `TermReference`, the `args` field is nullable. This allows us to distinguish between `-term` and `-term()`, which are functionally the same, but require different serialization.

spec/fluent.ebnf
```diff
@@ -56,32 +56,29 @@ block_placeable     ::= blank_block blank_inline? inline_placeable
  * syntax/abstract.mjs. */
 InlineExpression    ::= StringLiteral
                       | NumberLiteral
-                      | CallExpression
-                      | AttributeExpression
                       | VariantExpression
-                      | MessageReference
-                      | TermReference
-                      | VariableReference
+                      | ReferenceExpression
```
stasm: Would it make sense to fold `VariantExpression` into `ReferenceExpression`, even if it's deprecated? It might be closer to how the tooling parser will handle it.

syntax/ast.mjs
```diff
@@ -105,18 +105,21 @@ export class NumberLiteral extends Expression {
...
 }
 
 export class TermReference extends Expression {
-    constructor(id) {
+    constructor(id, attribute, args) {
         super();
         this.type = "TermReference";
         this.id = id;
+        this.attribute = attribute;
+        this.args = args;
```
stasm: Naming things... `attribute` or `attr` to make it consistent with `id` and `args`?

Pike: Full name sounds about right to me. `attr` is less obvious about the fact that this is exactly one or none.

spec/fluent.ebnf
Pike: I think it makes sense to have it explicit at this level.

PR #224, opened 2018-12-12
Update the syntax guide to Syntax 0.8

stasm:


stasm:
@flodolo, @Pike - Would you like to look at this update of the syntax guide, please? There's still work remaining to be done before 1.0 for sure, and I think we will progressively get there. In this PR my goal was to document the new features of Syntax 0.8. You can refer to the [changelog](https://github.com/projectfluent/fluent/blob/master/spec/CHANGELOG.md) for the complete list.

stasm:
If you're willing to clone my branch locally, you can run `npm run build:guide` and then use a tool like [`serve`](https://www.npmjs.com/package/serve) to serve it locally at localhost.

Or you can try unzipping this dist version which I just built: [guide.zip](https://github.com/projectfluent/fluent/files/2671304/guide.zip)

guide/multiline.md
```diff
@@ -0,0 +1,92 @@
...
+```
+
+This behavior also applies when the first line of a text block is indented
+relative to the following lines. This is the only case where a sequence of
+leading blank characters might be preserved as part of the text value, even
+if they're technically in the leading position.
+
+```
+multiline2 =
+        This message has four spaces of indent
```
flodolo: Is it worth to use the same visual clues used in multiline1 above? I find them very helpful.

guide/placeables.md
```diff
@@ -0,0 +1,32 @@
+# Placables
+
+Text in FTL may use special syntax to incorporate small pieces of
```
flodolo: FTL -> Fluent, since that seems the approach in the rest of the files.

guide/placeables.md
```diff
@@ -0,0 +1,32 @@
+# Placables
```
flodolo: Plac**e**able

guide/placeables.md
```diff
@@ -0,0 +1,32 @@
...
+
+Text in FTL may use special syntax to incorporate small pieces of
+programmable interface. Those pieces are denoted with curly braces `{` and `}`
+and are called placeables.
+
+It's common to use placeables to interpolate external
+[variables](variables.html) into the translation. Variables are provided by
+the developer and may change on runtime.
```
flodolo: may change on -> will be set?

guide/placeables.md
```diff
@@ -0,0 +1,32 @@
...
+
+```
+# $title is the title of the bookmark to remove.
```
flodolo: We should adopt the full syntax we are promoting in Gecko

```
# Variables:
#   $title (String) - Title of the bookmark to remove
```

guide/special.md
```diff
@@ -0,0 +1,115 @@
+# Special Characters
+
+In Fluent, text is the most important part of the file. As such, it doesn't
+have any special syntax: you can just write it without any delimiters (e.g.
+quotes). It also doesn't recognize any special escape sequences. Regular text
```
flodolo: Regular text: should we explicitly mention unicode somewhere?

guide/special.md
```diff
@@ -0,0 +1,115 @@
...
+should be enough to store the vast majority of translations. In rare cases
+when it's not, another type of text can be used: quoted text.
+
+## Quoted Text
+
+Quoted text uses double quotes as delimiters and cannot contain line breaks.
+It can be used similar to how other types of expressions can: inside
```
flodolo: Possible simplification: 

```
Like other types of expressions, it can be used inside a placeable (e.g. `{"abc"}`).
```

guide/special.md
```diff
@@ -0,0 +1,115 @@
...
+
+| Escape sequence | Meaning |
+|-----------------|---------|
+| `\"` | The literal double quote. |
+| `\uHHHH` | A Unicode character in the U+0000 to U+FFFF range. |
+| `\UHHHHHH` | Any Unicode character. |
+| `\\` | The literal backslash. |
+
+Escape sequences are rarely needed, but Fluent supports them for the sake of
+edge-cases. In real life application using the actual character in the
```
flodolo: nit: edge cases

guide/terms.md
```diff
@@ -9,44 +9,81 @@ consistently across the localization of the entire product.
...
+update-successul = { -brand-name } has been updated.
 ```
 
-## Terms and Variants
+## Parameterized Terms
 
-Values of terms may have multiple [variants](variants.html), including the
-default variant marked with an asterisk (`*`). For instance, variants may
-correspond to different grammatical cases.
+Term values follow the same rules as message values. They can be simple text,
```
flodolo: I'm quite lost in this paragraph. For example, I don't understand the `terms can only reference variables which have been explicitly defined` part. I'm likely oversimplifying things, but:

```
Term values follow the same rules as message values. They can be simple text, or
they can interpolate other expressions, including variables. However, while
messages will receive data for variables directly from the app, terms will
receive such data from the message referencing them. Such references take the
form of `-term(…)` where the variables available inside of the term are defined
between the parentheses, e.g. `-term(param: "value")`.
```

guide/terms.md
```diff
@@ -9,44 +9,81 @@ consistently across the localization of the entire product.
...
+```
+
+This pattern can be very useful for defining multiple _facets_ of the term,
+which can correspond to grammatical cases or other grammatical or stylistic
+properties of the language. In many inflected languages (e.g. German,
+Finnish, Hungarian, all Slavic languages), the *about* preposition governs
+the grammatical case of the complement. It might be accusative (German),
+ablative (Latin), or locative (Slavic languages). The grammatical cases can
+be defined as variants of the same tem and referred to via parameterization
+from other messages. This is what's going on in the `about` message above.
```
flodolo: going on -> happens

guide/text.md
```diff
@@ -1,16 +1,42 @@
-# Working With Text
+# Writing Text
 
+Fluent is a file format desgined for storing translations. In consequence,
```
flodolo: typo: designed

guide/text.md
```diff
@@ -1,16 +1,42 @@
...
+referred to as _patterns_.
+
+## Placeables
+
+In Fluent, text can interpolate values of other messages, as well as external
+data passed into the translation from the app. Use the curly braces to start
+and end interpolating another expression inside of a pattern:
+
+```
+# $title is the title of the bookmark to remove.
```
flodolo: Same note about comment format for variables.

guide/text.md
```diff
@@ -1,16 +1,42 @@
...
+remove-bookmark = Really remove { $title }?
```
flodolo: Maybe a more realistic `Are you sure you want to remove { $title }?`?

guide/text.md
```diff
@@ -1,16 +1,42 @@
...
-    app written entirely in <i>HTML5</i>.
-    It uses FTL to implement localization.
+
+Refer to the chapters about [placeables](placeables.html) for more
+information.
+
+## Multiline Text
+
+Text can span multiple lines. In such cases, Fluent will calculate the
+maximum indent common to _all_ indented lines and remove it from the final
```
flodolo: Same question: isn't it the minimum indent?

guide/multiline.md
```diff
@@ -0,0 +1,92 @@
+# Multiline Text
+
+Text can span multiple lines as long as it is indented by at least one space.
+Line breaks inside of multiline text are preserved in the final text value.
```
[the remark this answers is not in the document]
stasm: Good point. I'll try to rephrase this.

guide/multiline.md
```diff
@@ -0,0 +1,92 @@
...
+```
+# █ denotes the indent common to all lines (removed from the value).
+# · denotes the indent preserved in the final value.
+multiline1 =
+████This message has four spaces of indent
+████····on the second line of its value.
+```
+
+This behavior also applies when the first line of a text block is indented
+relative to the following lines. This is the only case where a sequence of
```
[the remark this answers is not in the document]
stasm: I've only ever seen it as _relative to_ and a quick dictionary search seems to confirm this: https://www.merriam-webster.com/dictionary/relative%20to. Then again, I'm not a native speaker, so I can be wrong.

guide/special.md
stasm: There's a whole paragraph about Unicode at the bottom of this file, but I think you're right: it's worth adding a note about it here as well.

guide/terms.md
stasm: This is indeed much better, thanks!

stasm:
Thank you for a in-depth review, @flodolo. I'll work on addressing your feedback later today.

guide/multiline.md
```diff
@@ -0,0 +1,92 @@
...
+```
+leading-spaces =     This message's value starts with the word "This".
+leading-lines =
+
+
+    This message's value starts with the word "This".
+    The blank lines under the identifier are ignored.
+```
+
+In multiline patterns, the maximum indent common to all indented lines is
```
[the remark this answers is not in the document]
flodolo: OK, I see what you mean. My brain expects *minimum*, because that's the shortest (minimum) among all the indentations. 

I also don't see that as 0, though; why do you? Are you thinking of this?

```
multi = Text can also span multiple lines
    as long as each new line is indented
    by at least one space.
```

For example, I don't expect it to be zero here:

```
block =
    Sometimes it's more readable to format
    multiline text as a "block", which means
    starting it on a new line. All lines must
    be indented by at least one space.
```

Going back to the problem, I still wonder if "minimum indent" would be clearer?

guide/multiline.md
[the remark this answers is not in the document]
flodolo: Good point, I didn't think of `to`. I stand corrected.

guide/special.md
```diff
@@ -0,0 +1,116 @@
...
+    To add an attribute to this messages, write
+    {".attr = Value"} on a new line.
+    .attr = An actual attribute (not part of the text value above)
+```
+
+## Escape Sequences
+
+Quoted text supports a small number of escape sequences. The backslash
+character (`\`) starts an escape sequence in quoted text. (In regular text,
+the backslash is just the literal character with no special meaning.)
```
flodolo: I would remove the parenthesis here.

guide/hello.md
```diff
@@ -1,19 +1,25 @@
...
-expressions to select the right variant depending on the circumstances.
+In its simplest form, a message has just a single text value; in the example
+above it's *Hello, world!*. The value begins at the first non-blank character
+after the `=` sign, but there are rare exceptions related to multiline text
+values. The next chapter ([Writing Text](text.html)) has all the details.
+
+Most of the messages you will work with in Fluent will look similar to
+the one above. Fluent has been designed to keep these simple translations
+simple. Sometimes, however, messages in Fluent can be more complex, have more
+than one value variant, one or more attributes, or use expressions to select
```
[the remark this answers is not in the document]
stasm: I like that, thanks.

guide/multiline.md
```diff
@@ -0,0 +1,92 @@
...
+    Sometimes it's more readable to format
+    multiline text as a "block", which means
+    starting it on a new line. All lines must
+    be indented by at least one space.
+```
+
+In almost all cases, patterns start at the first non-blank character and
+end at the last non-blank character. In other words, the leading and
+trailing blanks are ignored. There's one exception to this rule, due to
+another rule; see the `multiline2` message below.
```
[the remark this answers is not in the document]
stasm: I try to use _space_ to mean `U+0020` and and _blank_ to mean a space or a valid EOL.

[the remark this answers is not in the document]
stasm: I didn't call out tab in the guide. I'll add it at the top of the Multiline chapter, thanks.

PR #225, opened 2018-12-19
fixed typo in string example

msujaws:


stasm:
Thanks, @msujaws! _successful_ might be one of the hardest English words to spell for me :) I actually found another typo in the same file. I pushed a fix to your branch.

PR #241, opened 2019-02-19
Store the raw number in NumberLiteral

### LINKED ISSUE #239: Store number values as strings of digits
> I like to think about AST as a structured, semantic view of the source string.
>
> From that perspective, what if we stored a number as a slice `[start;end]` and parsed it in the resolver instead?
>
> My concern here is that storing numbers and parsing them differs between different programming languages Fluent may work in. Number is a low level primitive, but it still can differ in how it stores fractions, unsigned vs signed etc. By keeping AST free of that logic, we ensure that the AST is portable between the implementations irrelevant of how JS, Python or Rust treat numerical values.
>
> Not a blocker, just a suggestion.

_Originally posted by @zbraniecki in https://github.com/projectfluent/fluent.js/pull/342_

stasm:
One way to fix #239.

### DIFF
--- a/spec/fluent.ebnf
+++ b/spec/fluent.ebnf
@@ -57,7 +57,7 @@ InlineExpression    ::= StringLiteral
 
 /* Literals */
 StringLiteral       ::= "\"" quoted_char* "\""
-NumberLiteral       ::= "-"? digits ("." digits)?
+NumberLiteral       ::= integer fraction?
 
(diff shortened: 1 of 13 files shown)


stasm:
As per https://github.com/projectfluent/fluent/issues/239#issuecomment-466111172, this is a wontfix.

Issue #232, opened 2019-02-01
Store decimal fractions explicitly in NumberLiteral


stasm:
The `NumberLiteral` AST node currently has a `value` field. In most implementations it is stringly-typed in order to preserve the fractional part of the number, even if it's `0`. This is important because in some languages the plural rules change based on the mere presence of the fractional part, thus making it significant.

This approach has the benefit of being simple but it leaves the logic of converting the number literal into a `FluentNumber` type up to the implementation (e.g. `parseFloat`). Instead, we could clarify the expected behavior by parsing the number literal during parse-time. The resulting AST node could look as follows:

```
NumberLiteral {
    integer,
    fractional,
    places
}
```

`1.002` would parse as `{integer: 1, fractional: 2, places: 3}`. The `places` field is used to left-pad the fractional part with `0`.

Alternative approaches include:

 - Storing `1002` as an integer together with `places = 3` to move the decimal separator to the left.
- Storing the fractional part in reverse (`1.002` would parse as `1` and `200`) and assigning `places = null` when the number is a fraction-less integer.

Pike:
Another alternative is to store a float and a precision.

I'm expecting that the primary use-case here will be external variables and float plural rules? When would we have plural rules on number literals? I'm more expecting this to be a thing for FluentType instances passed into format()?

stasm:
> Another alternative is to store a float and a precision.

Ah, good point. I like that the best. I guess we should make the `FluentNumber` constructor to take these two arguments then: the float value and the precision. This is currently outside of the scope of the spec, however, since we don't have a reference resolver. We could write this down in text as a guidelines for implementing resolvers, however.

> I'm expecting that the primary use-case here will be external variables and float plural rules? When would we have plural rules on number literals? I'm more expecting this to be a thing for FluentType instances passed into format()?

You're right that this is mostly intended for `FluentTypes`. For `NumberLiterals` I can see it still be useful for exact matching on variant keys (`[1]` vs. `[1.0]`), and for examples (`{ 1.0 -> ...}`).

stasm:
This looked like a simple enough change to quickly try it out in code. See #235. I first wrote a test case which I'd like to merge regardless of whether this change lands. We don't currently have good testing coverage for numbers.

zbraniecki:
I like this change.

stasm:
[...]

PR #242, opened 2019-02-21
Revert changes to NumberLiteral from #232

### LINKED ISSUE #239: Store number values as strings of digits

stasm:
Fix #239. Revert changes from #232 which made `value` be a float, and added `precision`. Also rename `value` to `raw` for consistency with `StringLiteral`. I'll file another issue to discuss removing unescaping from the AST and moving it into the resolver.

### DIFF
--- a/spec/CHANGELOG.md
+++ b/spec/CHANGELOG.md
@@ -7,13 +7,6 @@
     The `VariantLists` and the `VariantExpression` syntax and AST nodes were
     deprecated in Syntax 0.9 and have now been removed.
 
-  - Store `NumberLiterals` as floats with precision. (#232)
-
-    The `NumberLiteral.value` field is now a float value rather than a string.
-    A new field called `precision` stores the number of decimal places used in
-    the literal, thus allowing the AST to preserve the precision used in the
-    source.
-
   - Rename `args` to `arguments`.
 
     The `args` field of `MessageReference`, `TermReference`,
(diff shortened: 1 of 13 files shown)


stasm:
@Pike, I'd like to land this separately from #243 to record the reverting of #232 explicitly in the history. To reduce the diff, perhaps I should drop the "rename value to raw" part, since we now we'll change it anyways in #243?

Pike:
Yeah, that might help, but would that make rebasing #246 hard?

stasm:
I intended to rebase/squash #246 anyways, and it wasn't hard at all :)

stasm:
I've updated this PR and I've renamed `NumberLiteral.raw` back to `value`, just as it was called before #232. This PR isn't a strict revert, however. I've simplified the `NumberLiteral` grammar production and I've  kept the test which I had added in #235.

PR #246, opened 2019-02-22
Add StringLiteral.parse(), NumberLiteral.parse()

### LINKED ISSUE #243: Remove unescaped string literals from the AST
With the reference resolver planned for the next quarter, I'm considering removing the unescaped version of string literals from the AST. The main motivation for adding it in #195 was to specify the unescaping mechanism precisely. We could still do that in a special-purpose reference function, with tests, for Fluent 1.0, and then move it into the resolver's spec in Q2.

This would be in line with how `NumberLiterals` only store the raw version of the number, too. See #242. The tooling use-cases that I'm aware of are interested in the raw string mostly, which allows running checks and serializing back to the same Fluent syntax. If the unescaping logic is provided as an independent function, other use-cases could benefit from it to cook the raw string and work with the unescaped one.

stasm:
Fix #243. The PR is based on #242.

@Pike, how does this look to you? I'll add tests for both `process` methods later today.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,9 +14,10 @@
     "generate:fixtures": "make -sC test/fixtures",
     "generate": "npm run --silent generate:ebnf && npm run --silent generate:fixtures",
     "lint": "eslint **/*.mjs",
-    "test:fixtures": "node --experimental-modules test/parser.mjs ./test/fixtures",
     "test:ebnf": "node --experimental-modules test/ebnf.mjs ./syntax/grammar.mjs ./spec/fluent.ebnf",
-    "test": "npm run --silent test:fixtures"
+    "test:fixtures": "node --experimental-modules test/parser.mjs ./test/fixtures",
+    "test:unit": "node --experimental-modules test/literals.mjs",
+    "test": "npm run --silent test:fixtures && npm run --silent test:unit"
   },
   "homepage": "https://projectfluent.org",
   "repository": {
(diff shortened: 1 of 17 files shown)


stasm:
Thanks for the review, it was very helpful. I've updated the PR to address your feedback. I'm still to write tests.

I considered a number of naming options for the processing method. I like your `parsed_value` suggestion, but in the end, I'm not sure if it works well with the return value which is an object. I think I'd like to keep it that way because it makes it easy to add more infromation about the literal in the future, if needed. `NumberLiteral` already uses it to return `precision` next to the `value`.

For this reason, I went for `Literal.parse()`. After all, for numbers, it parses both the value and the precision. I realize that _parse_ might be loaded in the context of a reference parser. I'm still open to other suggestions. Here are a few other ideas: `extract`, `evaluate`, `interpret`, `derive`, `read`.

Pike:
`parse` is fine, `extract` would also work for me. Less fond of the other options.

Pike:
Can you give me a hint on why `NumberLiteral.run` does something in tests? I can't find that ad-hoc.

stasm:
I apologize for not providing more context. My connectivity was poor last night. Is your question specifically about `NumberLiteral.run`, or about `StringLiteral.run` too?  Each production in `grammar.mjs` is a highly-specialized parser. The test take advantage of that and use `run` to parse some input, get a `Result`, and then `fold` it using a success function and an error function (similar to how `Promise.then` works).

PR #250, opened 2019-04-02
Update README.md

alabamenhu:
Add Perl 6 version to community implementations


stasm:
Thanks, and sorry this PR was open for so long.

alabamenhu:
No problem, although I'm going to need to submit a new PR soon because in the meantime Perl 6 got renamed to Raku :-)

It does remind me I need to put the finishing touches on implementation — because of the `NUMBER` and `DATETIME` functions being considered part of Fluent's core, I ended up needing to take a long pause to implement an ICU-like library first and that's just done.

PR #251, opened 2019-04-08
Syntax Compatibility Strategy document

stasm:
This document is based on my conversations with @flodolo and @Pike. The goal is to document the compatibility strategy for introducing and removing syntax features.

### DIFF
--- a/spec/compatibility.md
+++ b/spec/compatibility.md
@@ -13,16 +13,16 @@ primary purposes:
 ## Types of Changes
 
 _Backwards incompatible_ changes make old files not parse in the current
-parsers. Removing a feature which was previously available and reporting a
-`SyntaxError` instead is a backwards incompatible change.
+parsers. Removing a feature which was previously available and reporting
+a `SyntaxError` instead is a backwards incompatible change.
 
-_Extensions_ are changes which introduce syntax which was previously
-forbidden. They make current files with the extended syntax not parse in old
+_Extensions_ are changes which introduce syntax which was previously forbidden.
+Extensions make current files with the extended syntax not parse in old
 parsers. Old files continue to parse correctly in the current parsers.
 
 _Deprecations_ are non-breaking changes meant to encourage the use of more
 modern syntax and guarantee a transition period for migrating translations to
-the modern syntax. They may herald future removals of syntax features.
+the modern syntax. Deprecations herald future removals of syntax features.
 
 
 ## The 1.x
@@ -33,13 +33,16 @@ must maintain backwards compatibility with 1.0.
 The dot releases may introduce new syntax in a backwards compatible way. New
 features are only allowed as extensions to the 1.0 syntax and the AST.
 
-Extensions may be accompanied by deprecations. Deprecated syntax will
-continue to be supported throughout the 1.x lifetime. Parsers will report
-deprecations as `Annotations` in the `SyntaxNode.annotations` field.
-Authoring tools should encourage translators to use the new syntax. They are
-free to choose the length of the grace period by configuring which
-deprecations should be considered as errors, depending on the currently-used
-Fluent implementation in the target project.
+Extensions may be accompanied by deprecations. Deprecated syntax will continue
+to be supported throughout the 1.x lifetime. Parsers will report deprecations
+as annotations in the AST.
+
+Authoring tools should provide per-project settings for specifying the lower
+and upper bounds of accepted syntax versions, depending on the currently-used
+Fluent implementation in the target project. Syntax newer than the upper bound
+must not be allowed in the project's translations; syntax deprecated in
+versions older than the lower bound should be reported as warnings, and later
+on, as errors—depending on the grace period chosen for the project.
 
 
 ## The 2.0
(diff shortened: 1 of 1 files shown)


spec/compatibility.md
```diff
@@ -0,0 +1,73 @@
...
+parsers. Removing a feature which was previously available and reporting a
+`SyntaxError` instead is a backwards incompatible change.
+
+_Extensions_ are changes which introduce syntax which was previously
+forbidden. They make current files with the extended syntax not parse in old
+parsers. Old files continue to parse correctly in the current parsers.
+
+_Deprecations_ are non-breaking changes meant to encourage the use of more
+modern syntax and guarantee a transition period for migrating translations to
+the modern syntax. They may herald future removals of syntax features.
```
flodolo: I wonder if we should drop the "may", since removal is part of the overall strategy.

spec/compatibility.md
```diff
@@ -0,0 +1,73 @@
...
+
+The dot releases may introduce new syntax in a backwards compatible way. New
+features are only allowed as extensions to the 1.0 syntax and the AST.
+
+Extensions may be accompanied by deprecations. Deprecated syntax will
+continue to be supported throughout the 1.x lifetime. Parsers will report
+deprecations as `Annotations` in the `SyntaxNode.annotations` field.
+Authoring tools should encourage translators to use the new syntax. They are
+free to choose the length of the grace period by configuring which
+deprecations should be considered as errors, depending on the currently-used
```
flodolo: Maybe "as errors or simple warnings"?

spec/compatibility.md
```diff
@@ -0,0 +1,73 @@
...
+
+
+## The 2.0
+
+The 2.0 release is a breaking one. It removes the deprecated features of the
+syntax and keeps only the modern ones. Feature-wise, 2.0 is the last 1.x
+release with all deprecations removed.
+
+We’re developing Fluent with a future standard in mind. Fluent 2.0 will be a
+stepping stone towards it, achieving the following goals.
```
Pike: I'd like to rephrase this to make it clear that we don't intend to pull the rug under people's lives.

For the standards path, a wording like this might work: If Fluent gets adopted as part of another industry standard, it's expected that all deprecated features are stripped as part of the transition. In that sense, the deprecations inform projects if they're ready to migrate to the new standard.

For an evolution outside of a standards adoption, there might be a scenario in which all projects on all tools are on the a strict-1.x setting. At that point, it might make sense to define a 2.0 version.

Side thought, we might want to consider shipping a 1.x+1 together with a 2.1 release. That would add a feature in a backwards compatible way to 1.x and 2.x, and would allow shipping 2.0 w/out having to move the whole ecosystem off of 1.x.

I'm not sure this is helping or not.

spec/compatibility.md
```diff
@@ -0,0 +1,73 @@
...
+    complexity of the syntax.
+
+  - Finally, it will allow implementations to streamline their codebases,
+    which can increase their quality and give opportunities for new performance
+    optimizations.
+
+To help the users transition to 2.0, we’ll create a tool, `fluent1to2`, for
+performing a one-time migration from 1.x to 2.0. This tool would use the last
+stable version of the 1.x parser and would force the serialization of the
+deprecated syntax using the more modern alternatives, if possible.
```
Pike: I think we should strip this. This is something that may or may not work in our niche of the localization ecosystem, but for other toolchains, it might just not be a practical solution to serialize all their projects to disk, run a tool, and then parse all the files again.

What I do think we should do is highlight that deprecations may come with opportunities to automatic rewrites to alternative syntax. If so, deprecations should highlight those, and provide a proof-of-concept on how to do so automatically. That also highlights the impact these can have on bumping your warning and error levels on the 1.x path.

spec/compatibility.md
```diff
@@ -0,0 +1,73 @@
...
+
+
+## The 2.x and Beyond
+
+The dot releases in the 2.x lifetime will follow the same rules as 1.x. The
+3.0 will follow the same rules as 2.0.
```
Pike: I've been giving this more thought.

Fluent 2.0 might just not come at all. If we're successful, it might not be named Fluent after all. If it gets adopted, we should anticipate that the version compat story will be determined by that standards body at the time of adoption.

On the other hand, looking at a X.0 release with a bit of a German accent, an X.0 version is about removing bad ideas from Fluent. If we need to do this more than once, a lot of things should be looked at, including this document.

Realistically, I think this document will need to be reevaluated anyway if there's a  transition off of the Fluent 1.x release scheme.

I suggest to make that explicit.

stasm: Good point about the impact of Fluent being adopted by a standards organization. I'll add it.

I'd like to keep the mention of 3.0. The current strategy is pretty flexible from the standpoint of the spec authors (us). Its purpose is to set the expectations of the adopters. 3.0 _may_ be a breaking release, but it doesn't have to. It's easy to change the strategy to a stricter one after 2.x with no harm for the adopters.

spec/compatibility.md
stasm: I removed this passage. I'm not sure we need to mention automatic rewrites at all in this doc right now. A system for providing rewrite hints as part of reporting deprecations would indeed be nice; we can design it later and mention here, if needed.

spec/compatibility.md
stasm: It does help, but I'm not sure how to phrase this :) Let me push an update to the PR, and we can discuss this further.

spec/compatibility.md
```diff
@@ -0,0 +1,66 @@
...
+Extensions may be accompanied by deprecations. Deprecated syntax will continue
+to be supported throughout the 1.x lifetime. Parsers will report deprecations
+as annotations in the AST.
+
+Authoring tools should provide per-project settings for specifying the lower
+and upper bounds of accepted syntax versions, depending on the currently-used
+Fluent implementation in the target project. Syntax newer than the upper bound
+must not be allowed in the project's translations; syntax deprecated in
+versions older than the lower bound should be reported as warnings, and later
+on, as errors—depending on the grace period chosen for the project.
```
Pike: Thanks for the em dash here. I could go with spaces around it, but I know that's debated spelling.

spec/compatibility.md
```diff
@@ -0,0 +1,66 @@
+# Fluent Syntax Compatibility
+
+The goal of this document is to establish a shared vision for Fluent’s
+compatibility strategy after the 1.0 release. The vision will serve two
+primary purposes:
+
+  - (External) Manage early-adopters’ expectations of Fluent Syntax and
+    tooling.
+
+  - (Internal) Inform the triage during the 1.x lifecycle, 2.0 and beyond.
```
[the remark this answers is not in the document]
stasm: Thanks for the suggestion. I went with a slightly changed `(Internal) Contextualize feedback and inform design decisions during the 1.x lifecycle and beyond.`

stasm:
Thanks for the review, @Pike.

> I have one more general thought, but not pressing right now: Should we be explicit about not allowing insta-break changes? I can see people wondering "Dude, I can also just break stuff, what's then?", might be good to make that explicit?

Let's work on adding this in a follow-up. Adding it won't contradict the overall tone of the document.

PR #254, opened 2019-04-18
Use a placeholder email

missmatsuko:
In case that's someone's actual domain and/or email address

The example.com domain is expressly for this purpose:
http://example.com/


stasm:
Thank you!

PR #257, opened 2019-04-18
Duplicate switch case

MatheusRV:
If two cases in a switch statement have the same label, the second case will never be executed.

In JavaScript, cases in a switch statement can have arbitrary expressions as their labels. The interpreter does not check that these expressions are all different. At runtime, if two cases in a switch statement have the same label, the second case will never be executed. This most likely indicates a copy-paste error where the first case was copied and then not properly adjusted.


stasm:
Thanks for the PR! This must have been a copy&paste error. Well spotted!

PR #259, opened 2019-04-22
Add ordinal plural example to guide

eemeli:
As it's by no means obvious that `NUMBER()` may also be used to implement selectors on ordinal rather than cardinal plural categories, this PR adds a note about this to the syntax guide.

What's happening here is that the same options that are passed to `Intl.NumberFormat` are also passed to `Intl.PluralRules`, and NumberFormat ignores the `type` option. Admittedly there is a bit of a wtf stage that you go through if you start to think about this.

I'm not sure how JS-specific this functionality is.

### DIFF
--- a/guide/selectors.md
+++ b/guide/selectors.md
@@ -39,4 +39,17 @@ your-score =
     }
 ```
 
+Using formatting options also allows for selectors using ordinal rather than
+cardinal plurals:
+
+```
+your-rank = { NUMBER($pos, type: "ordinal") ->
+   [1] You finished first!
+   [one] You finished {$pos}st
+   [two] You finished {$pos}nd
+   [few] You finished {$pos}rd
+  *[other] You finished {$pos}th
+}
+```
+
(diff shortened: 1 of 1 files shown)


zbraniecki:
Maybe we should bring back explicit `PLURAL` for that? I'm concerned that `NUMBER` may be misleading.

eemeli:
I'd recommend keeping the current behaviour. Yes, it is admittedly a bit surprising, but it's not really misleading as such. It also conceptually matches the current special-casing of `NUMBER()` as a selector, as its options already influence the variant that's chosen. See for instance the `minFractionDigits` example immediately above my addition.

guide/selectors.md
```diff
@@ -39,4 +39,17 @@ your-score =
     }
 ```
 
+Using formatting options also allows for selectors using ordinal rather than
+cardinal plurals:
+
+```
+your-rank = { NUMBER($pos, type: "ordinal") ->
+   [1] You finished first!
+   [one] You finished {$pos}st
```
stasm: Great use of an explicit `[1]` variant combined with the generic `[one]`!

PR #261, opened 2019-04-27
Replaced MessageDateTimeArgument with FluentDateTime

danilobuerger:
See https://github.com/projectfluent/fluent.js/issues/222


stasm:
Thanks for the PR, and I'm sorry that it was open for so long!

PR #266, opened 2019-06-05
Use Prettier to auto-format code

stasm:


package.json
```diff
@@ -33,12 +34,12 @@
...
     "cli-color": "^1.2.0",
     "difflib": "^0.2.4",
-    "eslint": "^4.19.1",
     "fluent": "^0.6.4",
     "fluent-syntax": "^0.7.0",
     "gh-pages": "^1.1.0",
     "gitbook-cli": "^2.3.0",
-    "json-diff": "^0.5.2"
+    "json-diff": "^0.5.2",
+    "prettier": "^1.17.1"
```
Pike: This isn't recommended practice by prettier, https://prettier.io/docs/en/install.html.

stasm: Thanks. This sounds like a good practice, indeed.

stasm:
> I'm also not sure that using a tool that occasionally changes all files is the right thing for a specification repository. And pinning a monster of js dependencies for a long lived spec, even less so.

If it's going to speed up the development (and it will), I'm OK with it. Besides, the majority of the spec lives in the EBNF and the tests. For the resolver, there's going to be code, too, but Prettier is fairly orthodox wrt. to formatting, and I think it's OK to use it for our purposes.

Pike:
Stas and I talked about this a few weeks ago, and in particular for `grammar.js` decided to stay with explicitly formatted code.

The benefit of that is that that particular code is formatted to convey the intention of the Syntax spec, not just some code. Alternatives, sequences etc are semantics, not just lists that can be arbitrarily written.

Thus, I'll close this out.

PR #280, opened 2019-08-05
Complete test fixtures for invalid selector expressions, fixes #279

### LINKED ISSUE #279: Create test case to prevent nested selectors
We should audit the test coverage for 

https://github.com/projectfluent/fluent/blob/94dae4de9dcdbdb3a524191f6382b7886fb5bf31/syntax/abstract.js#L54-L61

Notably, I don't think we have a test to check for select expressions as selectors.

AFAICT, this also bleeds into our implementations.

Pike:
There were no tests for the full set of forbidden selector expressions.

That made it possible to pass the existing test suite while adding
`FTL.Placeable` to the list of allowed grammar productions in
`abstract.js`.
The test coverage is split between `select_expressions.ftl`, and
`reference_expressions.ftl`.

@zbraniecki, can you do the review duties here?

### DIFF
--- a/test/fixtures/select_expressions.ftl
+++ b/test/fixtures/select_expressions.ftl
@@ -21,6 +21,20 @@ invalid-selector-term-variant =
        *[key] value
     }
 
+# ERROR Nested expressions are not valid selectors
+invalid-selector-select-expression =
+    { { 3 } ->
+        *[key] default
+    }
+
+# ERROR Select expressions are not valid selectors
+invalid-selector-nested-expression =
+    { { $sel ->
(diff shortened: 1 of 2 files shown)


Pike:
It moves this from nothing to a syntax error. These tests just parsed and resolved in all syntax and runtime implementations, and in the js runtime, still do.

From a rough glance at the `@fluent/bundle` fixtures, it's also the only aspect of the spec where the `resource.js` parser actually yields messages for `foo-invalid-something` keys. Might be worth a follow-up issue on fluent.js.

PR #283, opened 2019-08-13
Add test validator.

Pike:
This test ensures that there's only one specification that's
allowed by the tests, at least for some well-known cases.

The cases include `not` <-> `maybe`, both of which can be empty,
and some modifications to the allowed selectors in abstract.js.

### DIFF
--- a/test/util.js
+++ b/test/util.js
@@ -37,6 +37,14 @@ export function readfile(path) {
     });
 }
 
+export function writefile(path, contents) {
+    return new Promise(function(resolve, reject) {
+        fs.writeFile(path, contents, function(err) {
+            return err ? reject(err) : resolve(path);
+        });
+    });
+}
+
 export const {diffString: diff} = json_diff;
 export const PASS = color.green("PASS");
 export const FAIL = color.red("FAIL");
--- a/test/validator.js
+++ b/test/validator.js
@@ -1,0 +1,113 @@
+import assert from "assert";
+import path from "path";
+import child_process from "child_process";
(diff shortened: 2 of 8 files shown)


Pike:
Talked with Zibi on what I'm looking for in the review here. Not just a code-review, but also an analysis if I missed something in what we can test in an automated fashion.

Thanks to @zoepage for the help in sketching this out. Also CC @SirNickolas as his questions triggered my attention here, and he probably cares that we're looking at this with more rigor.

test/util.js
```diff
@@ -37,6 +37,14 @@ export function readfile(path) {
     });
 }
 
+export function writefile(path, contents) {
+    return new Promise(function(resolve, reject) {
+        fs.writeFile(path, contents, function(err) {
+            return err ? reject(err) : resolve(path);
+        });
+    });
+}
```
zbraniecki: I wish we were already on node10+ and could use async here

test/validator.js
```diff
@@ -0,0 +1,113 @@
...
+            grammar = new_grammar;
+            changed = true;
+            console.log("new grammar");
+        }
+    }
+    await writefile(GRAMMAR, grammar);
+    return changed;
+}
+
+async function validate_selector() {
```
zbraniecki: It would be great to have a short comment at the top of this and previous functions which in describe tl;dr of what those tests are attempting to perform and test.

SirNickolas:
A few ideas on modifications. Don’t know how practical they are.

1. `maybe(parser)` → `(parser)`,
2. `repeat` ↔ `repeat1`,
3. `repeat` → `not`.

PR #286, opened 2019-08-20
Add more cases to validation.

Pike:
This is adding more options to more combinators.
There's one thing I didn't find out how to validate, and that
repeat1 should cover 1 and 2, and I don't think that's possible
to validate. I tried tracing the production calls in a different
patch, but that didn't return anything useful as it doesn't count
if the 2-match ends up in a parser error somewhere else in the
message.

@SirNickolas, can you give this a look? Also, Zibi?

### DIFF
--- a/lib/combinators.js
+++ b/lib/combinators.js
@@ -110,6 +110,12 @@ export function sequence(...parsers) {
         always([]));
 }
 
+// Test combinator to produce an empty list instead of
+// repeat or repeat1.
+export function repeat0(parser) {
+    return new Parser(stream => new Success([], stream));
+}
+
 export function repeat(parser) {
     return new Parser(stream =>
         parser
--- a/syntax/grammar.js
+++ b/syntax/grammar.js
@@ -2,11 +2,14 @@ import * as FTL from "./ast.js";
 import {list_into, into} from "./abstract.js";
 import {
     always, and, charset, defer, either, eof, maybe, not,
-    regex, repeat, repeat1, sequence, string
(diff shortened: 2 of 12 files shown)


SirNickolas:
Looks fine to me, thanks.

> There's one thing I didn't find out how to validate, and that
repeat1 should cover 1 and 2, and I don't think that's possible
to validate.

`repeat1` → `repeatExactly1`, where `repeatExactly1 = parser => parser.map(Array.of)`? This will give failures if a case with more than one element is not tested. Or have I misunderstood something?

One more thing that is not completely related to this PR. I suppose, in `select_expressions.ftl`, names for tests `invalid-selector-select-expression` and `invalid-selector-nested-expression` are mixed up.

Pike:
> Looks fine to me, thanks.
> 
> > There's one thing I didn't find out how to validate, and that
> > repeat1 should cover 1 and 2, and I don't think that's possible
> > to validate.
> 
> `repeat1` → `repeatExactly1`, where `repeatExactly1 = parser => parser.map(Array.of)`? This will give failures if a case with more than one element is not tested. Or have I misunderstood something?

I wish I could. I tried, and just have to admit that s.tas is the only one that actually understands the combinator code to the extent that one could write new ones.


> One more thing that is not completely related to this PR. I suppose, in `select_expressions.ftl`, names for tests `invalid-selector-select-expression` and `invalid-selector-nested-expression` are mixed up.

Right, fixing that up here, too.

PR #313, opened 2020-08-10
Reference tests for patterns starting with special chars

stasm:
A few tests for https://github.com/projectfluent/fluent.js/issues/512.

### DIFF
--- a/test/fixtures/special_chars.ftl
+++ b/test/fixtures/special_chars.ftl
@@ -1,0 +1,14 @@
+## OK
+
+bracket-inline = [Value]
+dot-inline = .Value
+star-inline = *Value
+
+## ERRORS
+
+bracket-newline =
+    [Value]
+dot-newline =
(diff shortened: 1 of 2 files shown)


stasm:
Ah, good point, I went by the test case from the original issue, but going by the EBNF, the test cases can be made more minimal.

stasm:
I think Travis is having a hiccup but I'll look again into the CI error tomorrow.

test/fixtures/special_chars.ftl
```diff
@@ -0,0 +1,12 @@
+bracket-inline = [Value]
```
Pike, line 1: You're missing the 

```properties
## OK

```

stasm: Grr, sorry, my bad. I removed it to re-trigger the build and forgot to re-add it :/

PR #318, opened 2020-12-29
Fix Polish declension of word "Firefox" in example

jrmajor:


zbraniecki:
Thanks for the PR!
We use the `ks` declension - `Firefoksa`, `Firefoksowi` - can you update the PR?

jrmajor:
@zbraniecki Thanks for your remark, I have fixed it. Is it ok now?

PR #324, opened 2021-01-24
Fix #323, add more line continuations to cr and crlf tests

### LINKED ISSUE #323: Add continuation with and without indention to cr and crlf tests
Referring back to https://github.com/projectfluent/fluent-rs/pull/217, it's apparently worthwhile to have tests for `cr` and `crlf` handling for empty lines as well as normal lines, with and without indent.

Looking at the tests there, we should have these upstream.

Pike:
I'm not happy with the `cr` test. It lumps all the combinations into the comments production, instead of actually testing `cr` handling in the various scenarios.

@zbraniecki, @stasm, any comments?

I'm entertaining the idea of splitting the `cr` test into a fixture for each block, but that's also very ad-hoc. Maybe with a trailing comment to explain what the test does and why.

### DIFF
--- a/test/fixtures/cr.ftl
+++ b/test/fixtures/cr.ftl
@@ -1,1 +1,1 @@
-### This entire file uses CR as EOL.err01 = Value 01err02 = Value 02err03 =    Value 03    Continued    .title = Titleerr04 = { "strerr05 = { $sel -> }
+### This entire file uses CR as EOL.err01 = Value 01err02 = Value 02err03 =    Value 03    Continued    and continued        and continued    .title = Titleerr04 = { "strerr05 = { $sel -> }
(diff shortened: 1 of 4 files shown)


zbraniecki:
Could you introduce "valid" lines as separators?

Pike:
Not sure if that makes the problem easier. We'd need to flag the file as binary with git attributes, and it'd reduce the number of editors that can actually edit the file. I'm already at the point where vscode doesn't work, and I tweak with vi and python byte strings.

Pike:
I'm not particularly fond of this patch, but I also don't have a better idea.

Pike:
@zbraniecki , any takes on this?

Can't flag anyone for review myself anymore, due to permissions.

zbraniecki:
@stasm - could you help us with your perspective on that?

PR #336, opened 2021-08-21
Add fluent-vue link to README

Demivan:
I'm not sure if it should be in this section. It is not an implementation as it just uses fluent.js.  
Maybe another section in README called "Integrations" is better. Official integrations `fluent-dom` and `fluent-react` could be added there.


spookylukey:
+1 to an integrations section, [django-ftl](https://github.com/django-ftl/django-ftl) could also go there.

alerque (approved):

eemeli (approved):

PR #348, opened 2022-05-29
Add Linguini

Mailaender:
Fluent.NET is dead, long live Linguini.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ This repository contains the reference implementation of the parser. Other imple
 We also know about the following community-driven implementations:
 
   - [`Fluent.Net`](https://github.com/blushingpenguin/Fluent.Net) by [@blushingpenguin](https://github.com/blushingpenguin). See [#93](https://github.com/projectfluent/fluent/issues/93) for more info.
+  - [`Linguini`](https://github.com/Ygg01/Linguini) a .NET library by [@Ygg01](https://github.com/Ygg01)
   - A Java/Kotlin implementation has been requested in [#158](https://github.com/projectfluent/fluent/issues/158).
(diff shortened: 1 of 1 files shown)


alerque:
Is Fluent.NET really dead? I think I would put both .NET projects on a line as alternatives, but maybe a 3 or 4 words that clarify the status or relation of the projects would be helpful.

Mailaender:
The author does not work for Mozilla anymore and ignores pull requests. It also has several fundamental flaws with weak types as this was a direct JS port and C# string culture being ignored, which is another reason for the rewrite. It would have saved me a lot of time if I knew this before.

README.md
```diff
@@ -44,6 +44,7 @@ This repository contains the reference implementation of the parser. Other imple
 We also know about the following community-driven implementations:
 
   - [`Fluent.Net`](https://github.com/blushingpenguin/Fluent.Net) by [@blushingpenguin](https://github.com/blushingpenguin). See [#93](https://github.com/projectfluent/fluent/issues/93) for more info.
+  - [`Linguini`](https://github.com/Ygg01/Linguini) by [@Ygg01](https://github.com/Ygg01)
```
rimas-kudelis, line 47: I think it would make sense to mention here that this is a .NET port as well, since it's not obvious from the name.

eemeli: Agreed. @Mailaender, if you could update the mention with such a clarifier, I'd be happy to accept this PR.

Ygg01:
I don't work for Mozilla, either, except spiritually.

Mailaender:
I can only say that we just ported to Linguini https://github.com/OpenRA/OpenRA/pull/20003 successfully to fix crash bugs from Fluent.NET

eemeli (approved):

Issue #256, opened 2019-04-18
Migrate away from gitbook


stasm:
The `gitbook` CLI tool [has been deprecated](https://github.com/GitbookIO/gitbook/commit/6c6ef7f4af32a2977e44dd23d3feb6ebf28970f4). Let's find an alternative, e.g. [`mdbook`](https://rust-lang-nursery.github.io/mdBook/).

flodolo:
Just stumbled upon this. We moved away from the the l10n docs, and it was pretty straightforward.

Hard to point to a single commit, but you can get an idea from
https://github.com/mozilla-l10n/documentation/blob/master/.travis.yml

One of the things we had to fix was absolute links
https://github.com/mozilla-l10n/documentation/commit/cd91558479462b4f35ce77baaacbf90e39492425

heyakyra:
Opened #362 but not really able to test since the deployment seems to be handled elsewhere

Also created https://github.com/heyakyra/fluent-companies if pepole know of organizations using Fluent

PR #362, opened 2023-12-28
Switch to mdBook from gitbook

heyakyra:
Addresses #256 

cc: @stasm @dminor @eemeli @gregtatum


flodolo:
Thanks for starting this. I've opened a separate PR with a slightly different approach (the same we use in other mozilla-l10n repositories).

To answer your question in the bug, I don't think there was an automated deployment step. It looks like it was done manually via `npm run`.

PR #363, opened 2024-01-11
Switch from gitbook to mdbook

### LINKED ISSUE #256: Migrate away from gitbook

flodolo:
Fixes #256

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -16,12 +16,16 @@ on:
 jobs:
   fetch:
     runs-on: ubuntu-latest
+    env:
+      MDBOOK_VERSION: v0.4.36
     steps:
       - name: Clone repository
         uses: actions/checkout@v4
-      - name: Install Rust packages
+      - name: Install mdbook
         run: |
-          curl --proto '=https' --tlsv1.2 -sSf https://rossmacarthur.github.io/install/crate.sh | bash -s -- --repo "rust-lang/mdBook" --bin mdbook --to ~/.cargo/bin
+          mkdir installed-bins
+          curl -Lf "https://github.com/rust-lang/mdBook/releases/download/$MDBOOK_VERSION/mdbook-$MDBOOK_VERSION-x86_64-unknown-linux-gnu.tar.gz" | tar -xz --directory=./installed-bins
+          echo `pwd`/installed-bins >> $GITHUB_PATH
       - name: Install gh-pages
         run: |
           npm install gh-pages@"~6.1.1"
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -6,6 +6,10 @@ on:
   push:
     branches:
       - master
+    paths:
+      - '.github/workflows/deploy.yml'
+      - 'book.toml'
+      - 'guide/**'
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,35 @@
+name: Deploy to GitHub Pages
+
+on:
+  # Triggers the workflow on push or pull request events but
+  # only for the default branch.
+  push:
+    branches:
+      - master
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  fetch:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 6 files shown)


flodolo:
The current rendering is available here: https://test.flod.org/tmp_guide/

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,39 @@
...
+
+jobs:
+  fetch:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v4
+      - name: Install Rust packages
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://rossmacarthur.github.io/install/crate.sh | bash -s -- --repo "rust-lang/mdBook" --bin mdbook --to ~/.cargo/bin
```
eemeli, line 24: This seems like a very surprising way to install mdBook.

The user guide includes a section on [Running `mdbook` in Continuous Integration](https://rust-lang.github.io/mdBook/continuous-integration.html), which suggests two different methods, in case we're not happy with just
```sh
cargo install mdbook
```
which should also "just work".

A decent indication that the recommended methods work is that they're used to build the [rust-lang docs site](https://github.com/rust-lang/cargo/blob/7bb7b539558dc88bea44cee4168b6269bf8177b0/.github/workflows/main.yml#L244-L248).

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,39 @@
...
+          npm install gh-pages@"~6.1.1"
+      - name: Build book
+        run: |
+            mdbook --version
+            mdbook build
+      - name: Set Git config
+        run: |
+          git config --global user.email "actions@users.noreply.github.com"
+          git config --global user.name 'Automation'
+          git remote set-url origin https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/$GITHUB_REPOSITORY
```
eemeli, line 36: Is this last line necessary? If we add a top-level config
```yaml
permissions:
  contents: write
```
would that not have the same effect, without requiring a `GITHUB_TOKEN` getting passed around?

.nojekyll
eemeli, line 1: Why is this added? AFAIK, there are no files or directories starting with an underscore in the mdbook results.

package.json
```diff
@@ -40,8 +40,7 @@
         "eslint": "^6.7.2",
         "@fluent/bundle": "^0.14.0",
         "@fluent/syntax": "^0.14.0",
-        "gh-pages": "^2.1.1",
-        "gitbook-cli": "^2.3.2",
+        "gh-pages": "^6.1.1",
```
eemeli, line 43: Should probably drop this and the `npm run deploy` script, if/as we're going to presume that the action will take care of that.

eemeli (changes requested):

.github/workflows/deploy.yml
flodolo: This is copied from our existing repositories, it looks like it became obsolete. As a disclaimer, this code came from someone very familiar with Rust and, at the time, the only alternative was compiling from source.

.github/workflows/deploy.yml
flodolo: Similarly, not sure this was available at the time. I'll test with the permissions bit.

.nojekyll
flodolo: We had [other errors](https://github.com/mozilla-l10n/documentation/pull/126#issuecomment-373093676) (again, a long time time ago). Since we don't use Jekyll, I tend to leave it there and disable it.

flodolo (commented):
I'll test the changes you suggested on another repo

.github/workflows/deploy.yml
flodolo: It doesn't work, I end up with an error
https://github.com/tschaub/gh-pages/issues/384

I'm guessing that I could create a PAT and use that, but I'd prefer avoiding it since this method works fine.

package.json
flodolo: Removed this and the deploy action, but I didn't run any npm command (as said, I'd prefer for someone who understands JS to do that)

eemeli (approved):

PR #365, opened 2024-01-22
Update node packages and use Node 20 LTS in CI

flodolo:


flodolo:
@eemeli I gave it a try, in case if it makes sense or you want to use it as a starting point. For some reasons tests fail on my machine, even with master.

no-idea-what-im-doing

That should also render obsolete a bunch of open pull requests from dependabot.

eemeli:
Closing in favour of #368.

Issue #349, opened 2022-08-04
Let's reduce the gap between Fluent and MessageFormat 2


eemeli:
While working on a JS runtime implementation for the upcoming [MessageFormat 2 syntax](https://github.com/unicode-org/message-format-wg/blob/develop/spec/syntax.md), I've also put together a two-way [Fluent/MF2 compatibility package](https://github.com/messageformat/messageformat/tree/master/packages/mf2-fluent/).

This has allowed me to identify a relatively small number of features that are likely to be available in MF2, but which are not currently available in Fluent. Some of these have previously been discussed here:
- #80
- #230 

Others I could not find prior discussions for:
- Allowing any literal values as variant keys
- Explicitly setting the available variables for message references, as is done for term references
- The representation of MF2 markup elements

In addition to the above, there are some Fluent features that are not supported in MF2:
- Giving more than one positional argument to a custom function
- Attributes on messages
- While message and term references are not supported in the MF2 syntax, they are achievable via custom MF2 functions

The work on MF2 is ongoing, but we're approaching a decent understanding of the eventual shape of messages in its 2.0 edition. It is therefore getting more and more interesting to minimize any friction between Fluent and MF2, and to think about releasing a 1.1 version of the Fluent syntax that would allow for the representation of all or nearly all possible MF2 messages, while continuing to parse all valid Fluent 1.0 messages with their current meaning.

In addition to desirable new Fluent language features, this would make it easier for tooling (such as [Pontoon](https://pontoon.mozilla.org/)) to provide a unified editing/translating experience for Fluent and MF2 messages with fewer gotchas and corner cases that depend on the underlying message format.

This issue is not intended to cover the discussion on all such differences, but to act as a collection point for enumerating them and considering whether it makes sense in the first place to develop Fluent in this direction. If that is desirable, separate conversations on each of those issues should follow.

Issue #208, opened 2018-11-07
Create a version-specific header for Fluent files


Pike:
For parsers, it's beneficial to be able to detect the version of a Fluent file.

Right now, we can use feature detection pretty well to distinguish 0.4 from 0.5-0.7 syntax.

Might make sense to add a header entry to denote syntax going forward.

I'd go with a semantic comment. `# @syntax 1.0` for example.

stasm:
Thanks for filing this. I've had the tab open for a few days now...

Were you thinking about making this part of the syntax (optional or required) or encouraging it as a convention? If it's the former, would it make sense to introduce a new comment type, e.g. `#!` for pragma comments?

Pike:
I'd stick to the realm of semantic comments, whether we require it or not. Effectively, one could treat `# @syntax` the same way as `#!`, in particular in a well-known location in the file. Thinking about runtime here specifically.

As a first step, I think we should stake the claim, aka, where to put it, what to put there.

I'd add it generously to documentation, examples, and serializers for this iteration.

stasm:
Is this more of a `docs`-label issue than `syntax`, then?

Pike:
Should we already start adding them to tests that use 0.8 features? Namely call expressions and then also 6-hex unicode escapes?
Also tests that check for non-`\` text, astrals, etc?

stasm:
Why not just do it to all tests in a single PR?

Pike:
We could do that, too.

stasm:
Let's move this out of the Syntax 0.8 milestone and label it with the `docs` label. I'm not sure we'll actually end up needing the version pragma in 0.x versions of the syntax, but it's something that we could consider for 1.0 and later.

eemeli:
It would probably not be a good idea to explicitly support `#!` pragma comments, as those would quite probably end up at the very beginning of the line, at which point it might get parsed as a shebang line. Or is that intentional, with usage a bit like this?
```
#!/usr/bin/env fluent-1.0
```

PR #379, opened 2026-01-02
Allow VariableReference as NamedArgument value

### LINKED ISSUE #230: Allow all inline expressions as named arguments
This was originally filed as #188 concerning `VariantExpressions` which was since removed from the Fluent Syntax. However, the same problem remains in named arguments to call expressions. 

Given a term defined as follows:

```properties
-thing = { $count ->
    *[one] thing
    [other] things
}
```

Should we allow this?

```properties
you-own = You have { $count ->
    *[one] a {-thing(count: $count)}
    [other] {-thing(count: $count)}
}.
```

Right now, only `NumberLiterals` and `StringLiterals` can be passed as named arguments, so the work around is to pass the names of requested variants explicitly:
```properties
you-own = You have { $count ->
    *[one] a {-thing(count: "one"}
    [other] {-thing(count: "other")}
}.
```

Or:

```properties
you-own = You have { $count ->
    *[one] a {-thing(count: 1}
    [other] {-thing(count: "other")}
}.
```

### LINKED ISSUE #325: We should add a fixture test that verifies that only strings or numbers can be values of NamedArgument
I just discovered that fluent-rs allows for any inline expression to be a value of a `NamedArgument` and no fixture checks that.

### LINKED ISSUE #337: reusing terms to create more complex terms
When terms inside other terms, there is a limitation around passing variables down to other terms, but I'm not sure about the nature of such limitation. Here is an example:

```
-bar = { $n -> 
    [0] some
    *[1] another
}
-foo = before { -bar(n: $x) } after
```

`-foo` is just an abstraction around `-bar`, delegating to it a piece of the message, but passing `$x` down is forbidden, with parsing error `E0014: Expected literal`.

cc @gonzalocordero

### LINKED ISSUE #369: Forward all variables to parameterized term
I'm building a React web app, and I'm looking for a way to make form validation error messages extensible and customizable. To do that, I'd like to forward all variables from one Fluent message to another message, similar to how you can spread function arguments into another function call in JS.

```
-formal-greeting = Good morning, { $user }!
-personal-greeting = Howdy, { $user }!

# `...arguments` passes all variables from the current context to the parameterized term.
# If the `greeting` message had a `$user` variable in scope, it will also be in scope for
# the terms.
greeting = { $mood ->
   [personal] { -personal-greeting(...arguments) }
  *[formal] { -formal-greeting(...arguments) }
}
```

# Use case

My app contains a form where the user can schedule events. The form contains two fields relevant for this use case:
1. An `<input type="email">` for inviting participants. Only members of my organization may be invited, which is validated by a `pattern="@example.org$"` validator.
2. An `<input type="date">` for the date. Events can only be scheduled within the next two weeks, but not on a weekend, which is enforced by 4 validators: `required`, `min={today}`, `max={today.add({ days: 14 })`, and a custom callback function that checks whether the user input is on a weekend.

I want to keep the error message language consistent with the remaining UI. For this reason I want to translate all error messages with Fluent, and not use the translations provided by the [constraint validation API](https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation#the_constraint_validation_api).

## Sharing common error messages

Most of my form fields use only default validators, like `required`, `min`, and so on. I defined a reusable Fluent message for all of the common errors (I borrowed the texts from the [Firefox translations](https://pontoon.mozilla.org/de/firefox/dom/chrome/dom/dom.properties/?search=FormValidation)).

```
default-errors = { $error -> 
   [ValueMissing] Please fill out this field.
   [NumberRangeUnderflow] Please select a value that is no less than { $min }.
   [NumberRangeOverflow] Please select a value that is no more than { $max }.
   [DateRangeUnderflow] Please select a value that is no earlier than { $date }.
   [DateRangeOverflow] Please select a value that is no later than { $date }.
   [TooShort] Please use at least { $minLength } characters (you are currently using { $length } characters).
   [TooLong] Please shorten this text to { $maxLength } characters or less (you are currently using { $length } characters).
   [PatternMismatch] Please match the requested format.
   [EmailMismatch] Please enter an email address.
   [UrlMismatch] Please enter a URL.
  *[other] Unknown error ({ $reason }).
}
```

Next, I implemented validation functions for these common errors.

```ts
type FieldValidationError =
  | { error: "ValueMissing" }
  | { error: "NumberRangeUnderflow"; min: number }
  | { error: "TooShort"; minLength: number; length: number }
  | ...

function required(value: string) {
  return !value ? { error: "ValueMissing" } : null;
}
function pattern(p: RegExp) {
  return (value: string) => !p.test(value) ? { error: "PatternMismatch" } : null;
}
function minLength(min: number) {
  return (value: string) => value.length < min ? { error: "TooShort", min, length: value.length } : null;
}
```

I wrote a hook that lets me define a form schema, and pass an array of validators to each field.

```ts
const today = Temporal.Now.plainDateISO();
const form = useForm({
  eventName: field<string>([required, minLength(3)]),
  participant: field<string>([required, email, pattern(/@example.com$/)]),
  date: field<Temporal.PlainDate>([required, min(today), max(today.add({ days: 14 })]),
});
```

When the user enters a value, the hook iterates over the validators for that field until one validator returns a non-null value. That error object is then passed to Fluent like this:

```ts
function translateError(error: FieldValidationError, bundle: FluentBundle, messageId = "common-errors") {
  const pattern = bundle.getMessage(messageId)!.value!;
  return bundle.formatPattern(pattern, error, []); // Pass the error object as `args`, exposing all error properties to the Fluent message.
}
```

For common error messages, this system works fine.

## Customizing the email pattern error message

So far, when a user tries to invite an email address that doesn't end in `@example.com`, they see the generic error message `Please match the requested format.` I want to override the error message for this one field only with a custom message. Here is how that could work with spread arguments:

```
participant-field-errors = { $error ->
   [PatternMismatch] You can only invite @example.com emails.
  *[other] { common-errors(...arguments) }
}
```

If the user input is rejected by the `pattern` validator, I can display a custom message. But if the user input is rejected by the `required` or `email` validator, I display the default error message.

## Implementing the "not-on-weekends" validator

Events mustn't be scheduled on weekends. This can be implemented with a custom validator.

```ts
function weekdaysOnly(value: Temporal.PlainDate) {
  return value.dayOfWeek >= 6 ? { error: "Weekend" } : null;
}
```

And here is the corresponding Fluent message:

```
date-field-errors = { $error ->
   [Weekend] Please select a weekday.
  *[other] { common-errors(...arguments) }
}
```

In this situation, it's important to pass _all_ variables from the current scope to the other message, because the `DateRangeUnderflow` message needs access to the `$date` variable.

# How to proceed?

I'm aware that Fluent is more or less [in maintenance mode at the moment](https://github.com/projectfluent/fluent/issues/358#issuecomment-1704071878). I'm also aware that even [passing single variables to terms](https://github.com/projectfluent/fluent/issues/230) is still an open discussion. Because of this, I don't expect a short-term decision or even implemention for this feature request.

My main goal is to point out this use case. Hopefully, I was able to demonstrate that this feature has a legitimate use case, and I hope it will be considered when Fluent development picks up again. I'm happy to contribute to the JS implementation and documentation when you're ready to move this forward.

eemeli:
- Closes #230
- Closes #325
- Closes #337
- See also #349
- Closes #369

This proposed syntax change would require the release of a new Fluent version, probably 1.1. It's forwards-incompatible, i.e. while all Fluent 1.0 content is valid Fluent 1.1, not all Fluent 1.1 content would be considered valid by a Fluent 1.0 processor.

Effectively, this would allow for usage like:
```fluent
-term = the { $foo } term
msg-1 = This message includes { -term(foo: $term) }.
msg-2 = This message formats { NUMBER($n, minimumFractionDigits: $fd) }.
```

Not all inline expression values are supported as named arguments, so e.g. `{ -term(foo: -bar) }` would continue to be a syntax error. My sense is that the strong, active desire is only for variable references, and that other uses are much more limited.

I'll be sumitting PRs for fluent.js and python-fluent to showcase the impact on implementations. From #325 I gather that fluent-rs might already support this, or that it's likely to be relatively straightforward to add there.

We'll need to consider carefully how to handle this in Pontoon, as we may need a per-project configuration of the supported Fluent version, or we need a solution for #208.

In Firefox, we'll need to review the capabilities of the fluent-rs version we're using.

### DIFF
--- a/spec/fluent.ebnf
+++ b/spec/fluent.ebnf
@@ -72,7 +72,7 @@ CallArguments       ::= blank? "(" blank? argument_list blank? ")"
 argument_list       ::= (Argument blank? "," blank?)* Argument?
 Argument            ::= NamedArgument
                       | InlineExpression
-NamedArgument       ::= Identifier blank? ":" blank? (StringLiteral | NumberLiteral)
+NamedArgument       ::= Identifier blank? ":" blank? (StringLiteral | NumberLiteral | VariableReference)
 
 /* Block Expressions */
 SelectExpression    ::= InlineExpression blank? "->" blank_inline? variant_list
--- a/guide/functions.md
+++ b/guide/functions.md
@@ -30,6 +30,10 @@ are only available to developers when they pre-format variables passed as
 arguments to translations (see [Partially-formatted
 variables](#partially-formatted-variables) below).
 
+Any inline expression may be used as a positional argument,
+but named argument values are limited to literal values in Fluent 1.0.
+In Fluent 1.1, a variable reference may also be used as a named argument value.
+
 ## Built-in Functions
 
 Built-in functions are very generic and should be applicable to any translation
--- a/guide/terms.md
+++ b/guide/terms.md
@@ -20,7 +20,8 @@ or they can interpolate other expressions, including variables. However,
 while messages receive data for variables directly from the app, terms
 receive such data from messages in which they are used. Such references take
 the form of `-term(…)` where the variables available inside of the term are
-defined between the parentheses, e.g. `-term(param: "value")`.
+defined between the parentheses, e.g. `-term(param: "value")` or
+`-term(param: $arg)` (starting from Fluent 1.1).
 
 ```
 # A contrived example to demonstrate how variables
--- a/syntax/grammar.js
+++ b/syntax/grammar.js
@@ -302,7 +302,8 @@ let NamedArgument = defer(() =>
         maybe(blank),
         either(
             StringLiteral,
-            NumberLiteral).abstract)
+            NumberLiteral,
+            VariableReference).abstract)
     .map(keep_abstract)
     .chain(list_into(FTL.NamedArgument)));
 
--- a/test/fixtures/call_expressions.ftl
+++ b/test/fixtures/call_expressions.ftl
@@ -19,6 +19,7 @@ positional-args = {FUN(1, "a", msg)}
 named-args = {FUN(x: 1, y: "Y")}
 dense-named-args = {FUN(x:1, y:"Y")}
 mixed-args = {FUN(1, "a", msg, x: 1, y: "Y")}
+variable-args = {FUN($foo, arg: $bar)}
 
 # ERROR Positional arg must not follow keyword args
 shuffled-args = {FUN(1, x: 1, "a", y: "Y", msg)}
--- a/test/fixtures/call_expressions.json
+++ b/test/fixtures/call_expressions.json
@@ -352,6 +352,58 @@
             "comment": null
         },
         {
+            "type": "Message",
+            "id": {
+                "type": "Identifier",
+                "name": "variable-args"
+            },
+            "value": {
+                "type": "Pattern",
+                "elements": [
+                    {
+                        "type": "Placeable",
+                        "expression": {
+                            "type": "FunctionReference",
+                            "id": {
+                                "type": "Identifier",
+                                "name": "FUN"
+                            },
+                            "arguments": {
+                                "type": "CallArguments",
+                                "positional": [
+                                    {
+                                        "type": "VariableReference",
+                                        "id": {
+                                            "type": "Identifier",
+                                            "name": "foo"
+                                        }
+                                    }
+                                ],
+                                "named": [
+                                    {
+                                        "type": "NamedArgument",
+                                        "name": {
+                                            "type": "Identifier",
+                                            "name": "arg"
+                                        },
+                                        "value": {
+                                            "type": "VariableReference",
+                                            "id": {
+                                                "type": "Identifier",
+                                                "name": "bar"
+                                            }
+                                        }
+                                    }
+                                ]
+                            }
+                        }
+                    }
+                ]
+            },
+            "attributes": [],
+            "comment": null
+        },
+        {
             "type": "Comment",
             "content": "ERROR Positional arg must not follow keyword args"
         },
--- a/test/fixtures/term_parameters.ftl
+++ b/test/fixtures/term_parameters.ftl
@@ -6,3 +6,4 @@ key01 = { -term }
 key02 = { -term () }
 key03 = { -term(arg: 1) }
 key04 = { -term("positional", narg1: 1, narg2: 2) }
+key05 = { -term(arg: $foo) }
--- a/test/fixtures/term_parameters.json
+++ b/test/fixtures/term_parameters.json
@@ -202,6 +202,51 @@
             },
             "attributes": [],
             "comment": null
+        },
+        {
+            "type": "Message",
+            "id": {
+                "type": "Identifier",
+                "name": "key05"
+            },
+            "value": {
+                "type": "Pattern",
+                "elements": [
+                    {
+                        "type": "Placeable",
+                        "expression": {
+                            "type": "TermReference",
+                            "id": {
+                                "type": "Identifier",
+                                "name": "term"
+                            },
+                            "attribute": null,
+                            "arguments": {
+                                "type": "CallArguments",
+                                "positional": [],
+                                "named": [
+                                    {
+                                        "type": "NamedArgument",
+                                        "name": {
+                                            "type": "Identifier",
+                                            "name": "arg"
+                                        },
+                                        "value": {
+                                            "type": "VariableReference",
+                                            "id": {
+                                                "type": "Identifier",
+                                                "name": "foo"
+                                            }
+                                        }
+                                    }
+                                ]
+                            }
+                        }
+                    }
+                ]
+            },
+            "attributes": [],
+            "comment": null
         }
     ]
 }


juliancoffee:
```
-creature-singular = { $creature ->
   [bonedragon] Bone Dragon
   [griffin] Griffin
   *[other] ??
 }
 
-creature-plural = { $creature ->
   [bonedragon] Bone Dragons
   [griffin] Griffins
   *[other] ??
}

# … Hundreds more ... for each

battle-log-attack-perish =
    { $attacker_count ->
        [one] The { -creature-singular(creature: $attacker-name) } does
       *[other] The { -creature-plural(creature: $attacker-name) } do
    } { $damage_points } damage. { $perish_count ->
        [one] One { -creature-plural(creature: $defender_name) } perishes.
       *[other] { $defender_count } { -creature-plural(creature: $defender_name) } perish.
    }
 ```
 
 I believe this also enables something like this from https://github.com/projectfluent/fluent/issues/80#issue-288643286, although in a bit roundabout way ^^'
 But with tooling supporting asymmetric terms (not all do, I know Weblate doesn't support that https://github.com/WeblateOrg/weblate/issues/9895), this should be doable for all the languages
 
 Resulting in:
 ```
 The Bone Dragon does 46 damage. 2 Griffins perish.
 ```
 ```
  The Griffin does 46 damage. 2 Griffins perish.
  ```
  But I haven't tested it yet, just judging from an example.
  
  To quote from the same issue:
  > I think we should do this, the use-cases look good enough. Localizers' life will be hard in these cases, but less hard than with the alternative.

**Comment by juliancoffee** [comment]
Or even something like that, from the same issue?
```fluent
-creature-genus = { $creature ->
    [elf] Maskulinum
    [fairy] Femininum
    *[other] ??
}

-creature-akkusative = { $creature ->
   [fairy] Fee
   [elf] Elfen
}

you-see =
    Du siehst { -creature-genus(creature: $object) ->
       *[Maskulinum] einen { -creature-akkusative(creature: $object) }
        [Femininum] eine { -creature-akkusative(creature: $object) }
        [Neutrum] ein { -creature-akkusative(creature: $object) }
    }.
```

Even if the original version is prettier, because you can pack everything together.
https://github.com/projectfluent/fluent/issues/80#issuecomment-359819560


I guess you can do it with selectors, too?

```fluent
-creature-db = { $creature ->
     [elf] { $attr -> {
         [singular] Elf
         [plural] Elfen
         [nominativ] Elf
         [akkusativ] Elfen
         [genus] Maskulinum
         *[other] ??
     }
     [fairy] { $attr -> {
         [singular] Fee
         [plural] Feen
         [nominativ] Fee
         [akkusativ] Fee
         [genus] Femininum
         *[other] ??
     }
}

you-see =
    Du siehst { -creature-db(creature: $object, attr: "genus") ->
       *[Maskulinum] einen { -creature-db(creature: $object, attr: "akkusativ" }
        [Femininum] eine { -creature-db(creature: $object, attr: "akkusativ" }
        [Neutrum] ein { -creature-db(creature: $object, attr: "akkusativ" }
    }.
```

eemeli:
> ```fluent
> -creature-singular = { $creature ->
>    [bonedragon] Bone Dragon
>    [griffin] Griffin
>    *[other] ??
>  }
>  
> -creature-plural = { $creature ->
>    [bonedragon] Bone Dragons
>    [griffin] Griffins
>    *[other] ??
> }
> 
> # … Hundreds more ... for each
> 
> battle-log-attack-perish =
>     { $attacker_count ->
>         [one] The { -creature-singular(creature: $attacker-name) } does
>        *[other] The { -creature-plural(creature: $attacker-name) } do
>     } { $damage_points } damage. { $perish_count ->
>         [one] One { -creature-plural(creature: $defender_name) } perishes.
>        *[other] { $defender_count } { -creature-plural(creature: $defender_name) } perish.
>     }
> ```

This doesn't work for languages with more than two plural forms, and it's mixing up terms that should be translated separately. You should instead do something like:

```fluent
creature-bonedragon = { $count ->
    [one] Bone Dragon
   *[other] Bone Dragons
 }
 
creature-griffin = { $count ->
    [one] Griffin
   *[other] Griffins
}

battle-log-attack-perish =
    { $attacker_count ->
        [one] The { $attacker } does { $damage_points } damage.
       *[other] The { $attacker } do { $damage_points } damage.
    } { $perish_count ->
        [1] One { $defender } perishes.
       *[other] { $defender_count } { $defender } perish.
    }
```
where `$attacker` and `$defender` are separately formatted `creature-*` strings.

juliancoffee:
> where $attacker and $defender are separately formatted creature-* strings.

Yep, that's what we do right now 😅 

As for that
> This doesn't work for languages with more than two plural forms, and it's mixing up terms that should be translated separately. You should instead do something like:

Yes, obviously. That's why I said that it would require allowing different languages define their own terms.
So it could be `-creature-one`, `-creature-few`, `-creature-other` for all number variants specific language needs. Even if round-trip via formatting is probably a better solution.

And I agree that it's rather clunky, that's why I showed second example with one big selector.
And example with cases and genus wouldn't work that way, because only *that* language can say what gender certain creature has.
Not to mention cases, which are wildly different in different languages.

I'm not sure what's the best way to handle more complex situations, especially if you need gender, plurality and grammatical case all at the same time.

But I still wanted to give examples what seems possible with this MR.
Sorry, if it was more of a noise ^^'

spec/fluent.ebnf
```diff
@@ -72,7 +72,7 @@ CallArguments       ::= blank? "(" blank? argument_list blank? ")"
 argument_list       ::= (Argument blank? "," blank?)* Argument?
 Argument            ::= NamedArgument
                       | InlineExpression
-NamedArgument       ::= Identifier blank? ":" blank? (StringLiteral | NumberLiteral)
+NamedArgument       ::= Identifier blank? ":" blank? (StringLiteral | NumberLiteral | VariableReference)
```
slatian, line 75: Is there any benefit to only allow this subset and not a full `InlineExpression` like for the positional `Argument`?

eemeli: The proposed is the smallest change required to address all of the use cases that have been raised.

As I mention above:
> Not all inline expression values are supported as named arguments, so e.g. `{ -term(foo: -bar) }` would continue to be a syntax error. My sense is that the strong, active desire is only for variable references, and that other uses are much more limited.

